PR #60, opened 2014-10-24
re-added DoxyGen support from old Makefile to CMakeLists

hholzgra:
slightly modified to be invoked by "make doxygen" and to put
results into Documentation/doxygen-html


hholzgra:
seems to be fixed

PR #63, opened 2014-10-28
make default build warning-free

hholzgra:
Fixed a lot of warnings (at least two might have been actual bugs),

and silenced a lot of "variable is never used", "(static) function is never used" and "assigned value is never used" warnings for now with extra CFLAGS

  -Wno-unused-variable -Wno-unused-but-set-variable -Wno-unused-function

as these don't seem to be critical / hiding actual bugs ...


hholzgra:
replaced by #72

PR #65, opened 2014-11-02
reverse diff order and use unified diff format in test

hholzgra:
so that additional output lines are marked as addition, not as deletion ... 
also use unified diff format (-u) to have diff context and not just the changed lines


markus456:
Please check if this is still relevant and if so, resubmit a pull request into the develop branch.

PR #56, opened 2014-10-08
Fixed ubuntu issues

y-trudeau:
Pretty interesting project, got some issues making .deb files, here're the fixes I found.  Any idea when this will be GA?

PR #71, opened 2015-02-13
Dont remove maxscale binary when stopping via init.d

zytek:


zytek:
Sorry for the double commit but had to do this via GitHub UI.

Nice surprise when evaluating MaxScale. :boom: 
FYI this was spotted on Ubuntu 12.04.

knowak:
Nice find... :trollface:

zytek:
Ouch, I see this was reported 4 months ago... :-1: 

https://github.com/mariadb-corporation/MaxScale/pull/56

markus456:
Closing, bug is fixed

zytek:
@markus456 I still see offending $PIDFILE variable in the repo, could you elaborate how it was fixed ? It works fine on Ubuntu Trusty, but removes binary on Ubuntu Precise

markus456:
It is fixed in the next release.

kolbe:
This wouldn't be a problem if you properly-quoted the variable expansions. You should *always* quote variable expansions in shell. So, this should be:

killproc -p "$PIDFILE" "$DAEMON"

PR #78, opened 2015-04-23
Make MaxScale handle zero-length files to aid bootstrapping.

sjmudd:
When bootstrapping a binlog router to start it needs to know the first
file to use. You can provide this information in the config file but
that will never be up to date, or you can copy a file to the binlog
server and start MaxScale. It will then carry on from the last file’s
current position.

What's better is to get the binlog server to pull the required data down
for you.

The binlog files have a 4-byte magic prefix so to start from the beginning
(position 4) you need to add these to an empty file if you do this by
hand. If you don’t then MaxScale will attempt to download from the
master at position 0 and the master will not accept this value. (This is
not apparent with a mysql client as `CHANGE MASTER TO ... MASTER_LOG_POS
= 0` 
triggers a write of the 4 magic bytes and then asks the master for
information from position 4 [not 0]).

This patch makes MaxScale behave similarly and allows you to only
need to touch the first binlog file to be downloaded for it to do the
right thing.

### DIFF
--- a/server/modules/routing/binlog/blr_file.c
+++ b/server/modules/routing/binlog/blr_file.c
@@ -165,6 +165,24 @@ blr_file_rotate(ROUTER_INSTANCE *router, char *file, uint64_t pos)
 
 
 /**
+ * binlog files need an initial 4 magic bytes at the start. blr_file_add_magic()
+ * adds them.
+ *
+ * @param router	The router instance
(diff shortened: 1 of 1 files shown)


sjmudd:
See: https://mariadb.atlassian.net/browse/MXS-115

sjmudd:
This is to confirm that this patch is provided under the BSD-new license.

PR #82, opened 2015-05-12
MXS-137: correct calculation of query length when length is >= 0x80

yhager:


yhager:
This contribution was submitted under the BSD-new license

PR #90, opened 2015-08-21
Fix init.d script in ubuntu

KawaiDesu:
Fix "/etc/init.d/maxscale: line 48: --user=maxscale: command not found" problem while using script from deb package (ubuntu 14.04).


tanji:
This is already fixed in the development branch. Please don't make pull requests vs. master.

essjayhch:
Yes but it's a bug fix. Something that stops your script from running. Last time I checked you can't download a development deb file.

KawaiDesu:
Which branch i should use to check if problem was fixed already and for pull requests?

markus456:
All pull requests should be made on develop branch.

PR #91, opened 2015-08-21
gw_str_xor null terminates the output

steved:


markus456:
Similar to other open source projects, the MariaDB Corporation needs to have a shared ownership of all code that is included in MaxScale. The easiest way to achieve this is by submitting your code under the BSD-new license. Are you OK with doing that?

MassimilianoPinto:
The null byte is not important: the data is used as it it is an we don't care about using it as a string.

Data is in binary format and may contain other nulls.

steved:
It's just a write overflow, the null byte isn't important. Submitting under BSD-new is fine, is there something specific I need to do?

PR #92, opened 2015-09-18
MXS-302: Allow local configuration overrides (e.g. to log to syslog)

sjmudd:
Please consider this or a similar patch to allow me to override the default init script behaviour.
Note: I'm not sure what template variable is correct for /etc/sysconfig, so that probably needs correcting.

This works on CentOS 6.

I use:

```
# override default maxscale startup options (to use syslog)
MAXSCALE_OPTIONS="--syslog=yes --maxscalelog=no"
```


sjmudd:
See: https://mariadb.atlassian.net/browse/MXS-302

markus456:
This has been implemented.

PR #93, opened 2015-10-19
Clarifies router_options parameter in documentation

swettk:


markus456:
The readwritesplit documentation has already been reworked in the develop branch. All pull requests should go to develop.

PR #102, opened 2015-12-14
Added new 'tpmfilter' filter plugin

dongyoungy:
Added a new filter plugin called 'tpmfilter' ('tpm' stands for 'transaction performance monitoring').
The filter automatically groups query statements into a transaction by detecting 'commit' or 'rollback'.
Each transaction is printed out with necessary information such as username, client, latency, etc., which enables future performance analysis of the DBMS using the log data generated by this filter.

Currently the plugin works with 'autocommit=0'. 

With 'autocommit=1', it should consider individual query as a single transaction, but I need to be revised on how the filter can figure out the value of 'autocommit' without the filter itself sending the query to the server, which I do not think is a correct use-case for MaxScale.

### DIFF
--- a/server/modules/filter/CMakeLists.txt
+++ b/server/modules/filter/CMakeLists.txt
@@ -46,6 +46,11 @@ target_link_libraries(namedserverfilter log_manager)
 set_target_properties(namedserverfilter PROPERTIES VERSION "1.1.0")
 install(TARGETS namedserverfilter DESTINATION ${MAXSCALE_LIBDIR})
 
+add_library(tpmfilter SHARED tpmfilter.c)
+target_link_libraries(tpmfilter log_manager)
+set_target_properties(tpmfilter PROPERTIES VERSION "1.0.0")
+install(TARGETS tpmfilter DESTINATION ${MAXSCALE_LIBDIR})
+
 
 if(BUILD_SLAVELAG)
   add_library(slavelag SHARED slavelag.c)
--- a/server/modules/filter/tpmfilter.c
+++ b/server/modules/filter/tpmfilter.c
@@ -60,7 +60,6 @@
 /** Defined in log_manager.cc */
 extern int            lm_enabled_logfiles_bitmask;
 extern size_t         log_ses_count[];
-extern __thread log_info_t tls_log_info;
 
 MODULE_INFO 	info = {
 	MODULE_API_FILTER,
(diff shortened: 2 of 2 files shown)


markus456:
Overall the filter looks good.
One thing that could be added is a small document in `Documentation/Filters/` which describes what it is for, how it works and how to use it. This will help people use it without having to look at the source code.

dongyoungy:
I have added the small document for the filter and also applied minor fixes.

The filter as it stands now is fully functional, but I have a question: 
It seems to me that it would be more than useful for the filter to print out information about a transaction has been executed on which server (e.g., if you have 'server1', 'server2' and a transaction is executed on 'server1', the filter prints 'server1' in the log for the transaction OR prints average latency from each server for a transaction). It is possible to extract such information given the current implementation of router and filter plugins? or should I write a new router that goes with the tpmfilter? GWBUF does not seem to have such information. Thank you.

markus456:
The document looks great, the usage example gives the user a good intro on how to use it and what to expect.

The server name where the query is routed isn't available in any convenient way. The router is the only module which knows where the query is going and where the response came from. To allow this information to be passed up to the filters, a minor modification to the filter interface's clientReply would be needed so that both the router and filter clientReply would have DCB as a parameter. I'm hoping we'll be able to implement this modification as soon as possible because this will allow more complex routing topologies in MaxScale.

markus456:

dongyoungy:
Hi Markus,

I am fine with contributing my code under the BSD-new license.

BTW, I have also implemented the router version of this plugin by modifying existing readconnroute router, which can now actually get the server name where the query is routed. I did not make a pull request for that because the implementation is kind of ad-hoc. I do not think that it follows the design principle of MaxScale (i.e. router logging stuff instead of filters). 

However, could you please look at the implementation if you have time? It would be great to hear your opinion. You can view the commit [here](https://github.com/dongyoungy/MaxScale/commit/2710b9d43ae82260fd8e0798a7fb759feb5171cc).

dongyoungy:
I am just checking on this pull request. What do you guys think of the router version of the plugin? I am willing to work to convert the proposed router plugin in the previous comment to the filter plugin by changing internals, if you guys can provide me a little bit of directions.

markus456:
The tpmfilter has been merged into the maxscale-experimental module package. We will most likely start distributing this experimental module package with the 2.1 release of MaxScale.

PR #104, opened 2016-03-09
Develop - adding Travis integration

rasmushoj:
Hi,

Here comes the need configuration + scripts to set up travis integration for MaxScale. The related files are the travis configuration file .travis.yml and the scripts build_maxscale.sh and download_mariadb.sh. 

In addition while configuring these I corrected a few things in other files:
1) cmake/FindMySQL.cmake, added PATHS to give the path to the embedded directory
2) etc/postinst.in, removed the copying of maxscale.conf and maxscale since that is already handled by cmake
3) query_classifier/qc_mysqlembedded/CMakeLists.txt, corrected variable for embedded library
4) server/core/monitor.c, removed mysql directory from include directive

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,31 @@
+language: c
+sudo: required
+dist: trusty
+compiler: gcc
+os: linux
+
+env:
+  global:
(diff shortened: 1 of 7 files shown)


markus456:
Squeezing the commits into one could be a good idea.  This would make the history cleaner.

rasmushoj:
The commits have now been squeezed.

rasmushoj:
I've reverted the changes in /etc/postinst.in and query_classifier/qc_mysqlembedded/CMakeLists.txt as you suggested, which made a lot of sense.

PR #112, opened 2016-11-02
Save time when domain name contains wildcard '%'

JindongChen:
This pull request is to saving time when MySQL user's domain name contains wildcard '%'.

I encounter an issue when using maxscale as binlog server. There are many users with wildcard domain name in my MySQL, so maxscale spent 5 minutes plus to get ip address of wildcard domain name at startup. As I know maxscale does not support wildcard hosts currently, so I add this patch to saving time for wildcard hosts.

### DIFF
--- a/server/core/gw_utils.c
+++ b/server/core/gw_utils.c
@@ -61,6 +61,12 @@ setipaddress(struct in_addr *a, char *p)
 
     hint.ai_socktype = SOCK_STREAM;
 
+    if (strchr(p, '%') != NULL)
+    {
+        MXS_INFO("Host %s contains wildcard, return", p);
+        return 0;
+    }
+
     /*
      * This is for the listening socket, matching INADDR_ANY only for now.
      * For future specific addresses bind, a dedicated routine woulbd be better


markus456:
I think this particular function is only called when the list of database users that exist in the backend database servers is being gathered. This means that it will only happen when a user fails to authenticate and MaxScale proceeds to check the existing users again. Regardless of the rarity of the function's use, it's a good improvement.

If you want this to be included in the next 2.0 release, please submit the pull request on the `2.0` branch.

JindongChen:
I'd like to contribute all my code in this pull request under the BSD-new license.

server/core/gw_utils.c
```diff
@@ -61,6 +61,12 @@ setipaddress(struct in_addr *a, char *p)
 
     hint.ai_socktype = SOCK_STREAM;
 
+    if (strchr(p, '%') != NULL)
+    {
+        MXS_ERROR("Host %s contains wildcard, return", p);
```
markus456: I think this message could be changed to `MXS_INFO` as it isn't really helpful to the end user. The INFO log level can be used when debugging connection or user problems and then this might be useful.

markus456 (changes requested):
Looks good, a minor change in the log level would be appropriate.

server/core/gw_utils.c
MassimilianoPinto: Tested with:

user@ foo%bar.bar
user@%.foo.bar

The patch can be applied but log message should be INFO:

MXS_ERROR ===> MXS_INFO

MassimilianoPinto (changes requested):
The modification has been succesfully tested.

Please change MXS_ERROR to MXS_INFO

JindongChen:
updated to INFO log level

MassimilianoPinto (approved):

markus456 (approved):

PR #114, opened 2016-12-02
Added a new 'dbseerroute' router plugin

dongyoungy:
Added a new router plugin 'dbseerroute', which is a modified version of readconnroute and logs transaction performance data that can be used by other open-source tools, such as (but not limited to) [DBSeer](http://dbseer.org). This plugin is very similar to 'tpmfilter', but unlike tpmfilter, it can print out information about a transaction being executed at which server, and also dynamically enable/disable logging via named pipe. Please refer to #102 for previous discussion.

The request does include little changes in core components of MaxScale (e.g., service.c, config.c, etc.). Please advise if such changes are inappropriate.

### DIFF
--- a/server/core/config.c
+++ b/server/core/config.c
@@ -1679,16 +1679,6 @@ process_config_update(CONFIG_CONTEXT *context)
                     char *version_string;
                     char *allow_localhost_match_wildcard_host;
 
-                    char *log_filename;
-                    char *log_delimiter;
-                    char *query_delimiter;
-                    char *named_pipe;
-
-                    log_filename = config_get_value(obj->parameters, "log_filename");
-                    log_delimiter = config_get_value(obj->parameters, "log_delimiter");
-                    query_delimiter = config_get_value(obj->parameters, "query_delimiter");
-                    named_pipe = config_get_value(obj->parameters, "named_pipe");
-
                     enable_root_user = config_get_value(obj->parameters, "enable_root_user");
                     connection_timeout = config_get_value(obj->parameters, "connection_timeout");
                     max_connections = config_get_value_string(obj->parameters, "max_connections");
@@ -1720,22 +1710,6 @@ process_config_update(CONFIG_CONTEXT *context)
                     if ((param = config_get_param(obj->parameters, "ignore_databases_regex")))
                     {
                         service_set_param_value(service, param, param->value, 0, STRING_TYPE);
-                    }
-                    if (log_filename)
-                    {
-                      serviceSetLogFilename(obj->element, log_filename);
-                    }
-                    if (log_delimiter)
-                    {
-                      serviceSetLogDelimiter(obj->element, log_delimiter);
-                    }
-                    if (query_delimiter)
-                    {
-                      serviceSetQueryDelimiter(obj->element, query_delimiter);
-                    }
-                    if (named_pipe)
-                    {
-                      serviceSetNamedPipe(obj->element, named_pipe);
                     }
 
                     if (version_string)
@@ -2719,28 +2693,6 @@ int create_new_service(CONFIG_CONTEXT *obj)
         service_set_param_value(obj->element, param, param->value, 0, STRING_TYPE);
     }
 
-    char* log_filename = config_get_value(obj->parameters, "log_filename");
-    char* log_delimiter = config_get_value(obj->parameters, "log_delimiter");
-    char* query_delimiter = config_get_value(obj->parameters, "query_delimiter");
-    char* named_pipe = config_get_value(obj->parameters, "named_pipe");
-
-    if (log_filename)
-    {
-      serviceSetLogFilename(obj->element, log_filename);
-    }
-    if (log_delimiter)
-    {
-      serviceSetLogDelimiter(obj->element, log_delimiter);
-    }
-    if (query_delimiter)
-    {
-      serviceSetQueryDelimiter(obj->element, query_delimiter);
-    }
-    if (named_pipe)
-    {
-      serviceSetNamedPipe(obj->element, named_pipe);
-    }
-
     char *version_string = config_get_value(obj->parameters, "version_string");
     if (version_string)
     {
@@ -2940,6 +2892,13 @@ int configure_new_service(CONFIG_CONTEXT *context, CONFIG_CONTEXT *obj)
     char *servers = config_get_value(obj->parameters, "servers");
     char *roptions = config_get_value(obj->parameters, "router_options");
     char *router = config_get_value(obj->parameters, "router");
+
+    /* options specific to dbseerroute. */
+    char *log_filename = config_get_value(obj->parameters, "log_filename");
+    char *log_delimiter = config_get_value(obj->parameters, "log_delimiter");
+    char *query_delimiter = config_get_value(obj->parameters, "query_delimiter");
+    char *named_pipe = config_get_value(obj->parameters, "named_pipe");
+
     SERVICE *service = obj->element;
 
     if (service)
@@ -2989,6 +2948,24 @@ int configure_new_service(CONFIG_CONTEXT *context, CONFIG_CONTEXT *obj)
             {
                 error_count++;
             }
+        }
+
+        // set dbseerroute specific options
+        if (log_filename)
+        {
+            serviceSetLogFilename(service, log_filename);
+        }
+        if (query_delimiter)
+        {
+            serviceSetQueryDelimiter(service, query_delimiter);
+        }
+        if (log_delimiter)
+        {
+            serviceSetLogDelimiter(service, log_delimiter);
+        }
+        if (named_pipe)
+        {
+            serviceSetNamedPipe(service, named_pipe);
         }
     }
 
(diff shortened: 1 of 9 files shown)


server/modules/routing/dbseerroute/readconnection.h
```diff
@@ -0,0 +1,94 @@
+#pragma once
+#ifndef _READCONNECTION_H
```
markus456: The `develop` branch currently only uses `#pragma once` instead of both pragma once and header guards. You can remove the header guards and just leave `#pragma once`.

server/core/config.c
```diff
@@ -1675,8 +1679,17 @@ process_config_update(CONFIG_CONTEXT *context)
                     char *version_string;
                     char *allow_localhost_match_wildcard_host;
 
-                    enable_root_user = config_get_value(obj->parameters, "enable_root_user");
+                    char *log_filename;
+                    char *log_delimiter;
```
markus456: Since these are only used by the dbseerrouter, they should be passed as values for `router_options`. We're trying to move away from a list of pre-defined parameters and move to a more dynamic configuration processing.

If you need to change these at runtime, we've just added a way for modules to declare custom commands for MaxAdmin. See modulecmd.h and the related functions. The dbfwfilter is one which implements a few extra commands in the `ModuleInit` function and you can use that as an example.

server/modules/routing/dbseerroute/readconnection.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Change Date: 2019-07-01
+ *
+ * On the date above, in accordance with the Business Source License, use
+ * of this software will be governed by version 2 or later of the General
+ * Public License.
+ */
+
+/**
+ * @file readconnection.h - The read connection balancing query module heder file
```
markus456: These comments refer to the readconnroute header. It would be better to just remove them since this is a separate header.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+            }
+            else if (!strcasecmp(options[i], "ndb"))
+            {
+                inst->bitmask |= (SERVER_NDB);
+                inst->bitvalue |= SERVER_NDB;
+            }
+            else
+            {
+                MXS_WARNING("Unsupported router "
+                            "option \'%s\' for readconnroute. "
```
markus456: The error message mentions readconnroute even though this is a different module.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+        inst->bitvalue |= SERVER_RUNNING;
+    }
+
+    /*
+     * check named pipe.
+     */
+    inst->log_enabled = false;
+    if ((named_pipe = serviceGetNamedPipe(service)) != NULL)
+    {
+        inst->named_pipe = strdup(named_pipe);
```
markus456: Use MXS_STRDUP here.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+        {
+            // if exists, check if it is a named pipe.
+            struct stat st;
+            ret = stat(inst->named_pipe, &st);
+
+            // check whether the file is named pipe.
+            if (ret == -1 && errno != ENOENT)
+            {
+                MXS_ERROR("stat() failed on named pipe: %s", strerror(errno));
+                free(inst);
```
markus456: Use MXS_FREE here.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+
+    /*
+     * process logging options.
+     */
+    if ((log_filename = serviceGetLogFilename(service)) != NULL)
+    {
+        MXS_FREE(inst->log_filename);
+        inst->log_filename = strdup(log_filename);
+        /* set default log and query delimiters */
+        inst->log_delimiter = strdup(":::");
```
markus456: You could assign the default values to `const char*` pointers and then replace them with the custom ones so that you'd only have to do one memory allocation per variable.

In addition to this, MaxScale currently checks the return values of memory allocations. This means that either the code should check the return values for MXS_STRDUP etc. or use one of the _A suffix macros that abort on memory allocation error. I think you can just use MXS_STRDUP_A here without checking for return values as this is part of the system startup code.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+            MXS_ERROR("Failed to open a log file for dbseerroute router.");
+            free(inst);
+            return NULL;
+        }
+    }
+
+    /*
+     * Launch a thread that checks the named pipe.
+     */
+    pthread_t tid;
```
markus456: The `include/maxscale/thread.h` header has some functions that start threads for you. This removes the need to directly interact with the threading library.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+            free(client_rses);
+            return NULL;
+        }
+    }
+
+    /*
+     * Set up for logging if log filename is specified.
+     */
+    client_rses->max_sql_size = 4 * 1024;
+    client_rses->sql = (char*)malloc(client_rses->max_sql_size);
```
markus456: Use MXS_MALLOC.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+        /* check for commit and rollback */
+        if (query_len > 5)
+        {
+             int query_size = strlen(ptr)+1;
+             char *buf = router_cli_ses->buf;
+             for (int i = 0; i < query_size && i < buf_size; ++i)
+             {
+                 buf[i] = tolower(ptr[i]);
+             }
+             if (strncmp(buf, "commit", 6) == 0)
```
markus456: You can use the query classifier for this. See `qc_get_type` in `query_classifier.h`.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+                 goto return_rc;
+            }
+            /* double buffer size until the buffer fits the query.*/
+            while (len > new_sql_size)
+            {
+                new_sql_size *= 2;
+            }
+            if (new_sql_size > router_cli_ses->max_sql_size)
+            {
+                char* new_sql = (char*)malloc(new_sql_size);
```
markus456: You can use MXS_REALLOC here.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+            timersub(&tv, &(router_cli_ses->current_start), &diff);
+
+            /* get latency.*/
+            uint64_t millis = (diff.tv_sec * (uint64_t)1000 + diff.tv_usec / 1000);
+            /* get timestamp.*/
+            uint64_t timestamp = (tv.tv_sec + (tv.tv_usec / (1000*1000)));
+            *(router_cli_ses->sql + router_cli_ses->sql_index) = '\0';
+
+            char *delimiter = inst->log_delimiter;
+            char* server_uniquename = router_cli_ses->backend->server->unique_name;
```
markus456, line 892: I think this could be done in the tpmfilter as well if we add a `SERVER` pointer to the GWBUF which tells where this buffer came from. It would allow filters to track the origin of a particular buffer.

server/modules/routing/dbseerroute/dbseerroute.c
```diff
@@ -0,0 +1,1181 @@
...
+ */
+static void rses_end_locked_router_action(ROUTER_CLIENT_SES* rses)
+{
+    CHK_CLIENT_RSES(rses);
+    spinlock_release(&rses->rses_lock);
+}
+
+static uint64_t getCapabilities(void)
+{
+    return RCAP_TYPE_NONE;
```
markus456: This should return the statement tracking capability defined in `routing.h` if complete SQL statements are needed.

markus456 (changes requested):
This looks quite interesting. A few changes are required so that it conforms to current code guidelines.

I think it would be better to alter MaxScale so that the tpmfilter could do this instead of writing a new router module that fixes problems in MaxScale. We are OK with changes that would change the MaxScale core as long as they make sense.

dongyoungy:
Thank you for the comments/suggestions.
I have made changes as they have been suggested, except for 'router_options', as they are already used by readconnroute (dbseerroute is basically readconnroute + logging feature). I have added separate options in the _configure_new_service_ function instead, while removing them from _process_config_update_ as then do not need to be updated on the fly.

I have also made changes in tpmfilter to use MXS_FREE and MXS_STRDUP_A instead of vanilla C functions.

I am not sure how to pass the required information so that tpmfilter can do the same job as the dbseerroute at this point. If you have any suggestions, I am more than happy to work on that as well.

dongyoungy:
I have fixed my stupid typo so that it now passes all the checks. Thanks!

markus456:
Pull request #116 was merged, closing this one.

PR #116, opened 2016-12-14
Improved TPM filter plugin

dongyoungy:
From the suggestions in the pull request #114, I have updated the TPM filter so that it can now log information about a transaction being executed at which server, and also dynamically enable/disable logging via named pipe. As discussed in #114, GWBUF now has a SERVER pointer for the server where the original transaction has been executed. the pointer gets assigned in gw_read_and_write() under mysql_backend.c.

I think this approach is better than new dbseerroute plugin in #114. If this one gets accepted by maintainers, #114 should be disregarded and closed without merge.


include/maxscale/buffer.h
```diff
@@ -145,6 +147,7 @@ typedef struct gwbuf
     gwbuf_type_t    gwbuf_type; /*< buffer's data type information */
     HINT            *hint;  /*< Hint data for this buffer */
     BUF_PROPERTY    *properties; /*< Buffer properties */
+    struct server   *server;
```
markus456: A small inline comment about what is struct is would be good. It helps when the documentation for member variables gives a small hint as to what it's going to be used for.

server/modules/filter/tpmfilter/tpmfilter.c
```diff
@@ -259,30 +345,30 @@ newSession(FILTER *instance, SESSION *session)
...
 
-    if ((my_session = calloc(1, sizeof(TPM_SESSION))) != NULL)
+    if ((my_session = MXS_CALLOC(1, sizeof(TPM_SESSION))) != NULL)
     {
         atomic_add(&my_instance->sessions, 1);
 
-        my_session->max_sql_size = 4 * 1024; // default max query size of 4k.
-        my_session->sql = (char*)malloc(my_session->max_sql_size);
+        my_session->max_sql_size = default_sql_size; // default max query size of 4k.
+        my_session->sql = (char*)MXS_MALLOC(my_session->max_sql_size);
```
markus456: You could use `MXS_CALLOC` here.

server/modules/protocol/MySQL/MySQLBackend/mysql_backend.c
```diff
@@ -895,6 +895,7 @@ gw_read_and_write(DCB *dcb)
         if (session_ok_to_route(dcb))
         {
             gwbuf_set_type(stmt, GWBUF_TYPE_MYSQL);
+            stmt->server = dcb->server;
```
markus456: better place to assign this pointer would be in `dcb_read` where the buffer is allocated. The DCBs have a SERVER pointer which could always be assigned to the buffer since the client DCB's SERVER pointer is NULL.

Look at line 870 in `server/core/dcb.c`, I think this would be a good place to do it.

markus456 (changes requested):
This is a nice addition to the tpmfilter and it seems mostly OK. A few minor things that could be changed but apart from that, it's OK.

server/core/dcb.c
```diff
@@ -879,6 +879,8 @@ int dcb_read(DCB   *dcb,
                 /* </editor-fold> */
                 /*< Append read data to the gwbuf */
                 *head = gwbuf_append(*head, buffer);
+                /*< Assign the target server for the gwbuf */
+                buffer->server = dcb->server;
```
markus456, line 881: This would be better to do before the `gwbuf_append` that any changes to the function would not affect this code.

markus456 (approved):
One very minor comment about the change but overall, I'd say it's ready to be merged.

markus456 (approved):

markus456:

dongyoungy:
I am contributing my code in this whole pull request under the BSD-new license.

PR #118, opened 2017-02-08
Zendesk's Maxwell Compatibility

adaszko:
Greetings!

Here are the changes I needed to apply in order to get Maxwell working with MaxScale.

They're exclusively about handling few SQL queries that Maxwell expects MaxScale to handle.  I leveraged the existing infrastructure to issue the queries known a priori, on MaxScale's startup, cache the responses from master and then replay the responses to Maxwell when it asks for them.

This PR may have few rough edges.  The worst part being hard-coded comment in a SQL query containing mysql-connector-j version string (!).  The reason I create the PR so early is I want to get the discussion started whether there's any intention of merging these changes.  For example, if the plan is to wait on https://jira.mariadb.org/browse/MXS-448 being implemented instead, there's no point in me working on this any further.

### DIFF
--- a/server/modules/routing/binlogrouter/blr.h
+++ b/server/modules/routing/binlogrouter/blr.h
@@ -496,6 +496,7 @@ typedef struct
     GWBUF           *mariadb10;         /*< set @mariadb_slave_capability */
     GWBUF           *server_vars;       /*< MySQL Connector master server variables */
     GWBUF           *results_charset;   /*< SET character_set_results = NULL */
+    GWBUF           *sql_mode;          /*< SET sql_mode='...' */
     uint8_t         *fde_event;         /*< Format Description Event */
     int             fde_len;            /*< Length of fde_event */
 } MASTER_RESPONSES;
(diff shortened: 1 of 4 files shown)


markus456:
Better handling of the queries is intended to be implemented but we're not certain as to how and when it'll be implemented.

I took a look at the changes and they don't seem to introduce any big changes that could alter how the binlogrouter currently works. We can run tests on the branch to verify this and if no changes occur in the normal use of binlogrouter, I'd say that we can merge this pull request. Before we merge it, it would be good to remove the explicit connector version string and handle it so that all versions would work.

MassimilianoPinto:
Hi,

**SELECT @@global.binlog_checksum**
should be the same we already save in `router->saved_master.chksum2` via

SET @master_binlog_checksum = @@global.binlog_checksum")
and
SELECT @master_binlog_checksum"

So you can just add in blr_slave.c:
```
else if (strcasecmp(word, "@master_binlog_checksum") == 0) || (strcasecmp(word, "@@global.binlog_checksum") == 0))
```

and remove the field binlog_checksum in MASTER_RESPONSES struct.

**SET sql_mode='NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES';**
**SET character_set_results = NULL;**

These SET can return either an error or OK packet (without any status message)

If this is always true, OK without messages, you can remove these fields in MASTER_RESPONSES struct: results_charset and sql_mode because you don't need to store the results (blr_master.c)

 In blr_slave.c just return blr_slave_send_ok(router, slave); when those SET are requested.

SET commands can alter the MaxScale to Master registration phase.

I suggest to create an option,say 'maxwell_compatibility=0|1, where default is 0.

This way alle the new queries and SET are sent only in the option is 1.

Of course the **mysql-connector-j version string** must be handled some how: say an option that specify a file to read or just a new router_option.

Another improvement I can see is:
instead of using SHOW VARIABLES LIKE 'log_bin', SHOW VARIABLES LIKE 'binlog_format',SHOW VARIABLES LIKE 'binlog_row_image'

SELECT @@global.log_bin, @@global.binlog_format, @@global.binlog_row_image;
then you could parse the result with blr_extract_column(buf, 1); (1,2,3)
This way you can set the router->saved_master.xyz response (only one) and you don't add the BLRM_LOG_BIN , BLRM_BINLOG_FORMAT, BLRM_BINLOG_ROW_IMAGE simplifying the registration process.

You can use only one state, say BLR_LOG_BIN_DETAILS and then create variables from one saved reply:

```
char *var1 = blr_extract_column(router->saved_master.selectver, 1);
blr_slave_send_var_value(router, slave, "@@var1", var1 == NULL ? "" : var1, BLR_TYPE_STRING);
```

MassimilianoPinto:
Hi, could you please add a new pull request or update the previous one?

adaszko:
> Hi, could you please add a new pull request or update the previous one?

Sure, I'm currently testing code that addresses all your remarks.  I'll submit it as soon as I'm done with it.

adaszko:
Opened #119

PR #119, opened 2017-02-10
Zendesk's Maxwell Compatibility

adaszko:
Addressed feedback from #118.

### DIFF
--- a/server/modules/routing/binlogrouter/blr.c
+++ b/server/modules/routing/binlogrouter/blr.c
@@ -185,6 +185,7 @@ MXS_MODULE* MXS_CREATE_MODULE()
             {"master_version", MXS_MODULE_PARAM_STRING},
             {"master_hostname", MXS_MODULE_PARAM_STRING},
             {"mariadb10-compatibility", MXS_MODULE_PARAM_BOOL, "false"},
+            {"maxwell-compatibility", MXS_MODULE_PARAM_BOOL, "false"},
             {"filestem", MXS_MODULE_PARAM_STRING, BINLOG_NAME_ROOT},
             {"file", MXS_MODULE_PARAM_COUNT, "1"},
             {"transaction_safety", MXS_MODULE_PARAM_BOOL, "false"},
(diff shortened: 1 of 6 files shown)


MassimilianoPinto:
Hi,

1)
**router->maxwell_compat** should be used for skipping "SET character_set_results = NULL" as well.

I would suggest to use **router->maxwell_compat** to skip all the "maxwell" queries.

2)
If possible don't save the output of SET (that should be OK packet without messages) and reply to client with "blr_slave_send_ok".

3)
I haven't tested the code yet but I see:
```
+    else if (strcmp(query_text, maxwell_binlog_row_image_query) == 0)
 +    {
 +        char *binlog_row_image = blr_extract_column(router->saved_master.binlog_vars, 2);
```

The number should be 3, isn't it ?

adaszko:
The column numbering starts at 0 or at 1?  It's the third column that I want to extract there.

MassimilianoPinto:
The column number starts at 1.

In the code I see:

```
char *log_bin = blr_extract_column(router->saved_master.binlog_vars, 1);
char *binlog_format = blr_extract_column(router->saved_master.binlog_vars, 2);
char *binlog_row_image = blr_extract_column(router->saved_master.binlog_vars, 2);
```

You need to change the last one from 2 to 3

adaszko:
I'm applying all your suggestions.  One question though:

> I would suggest to use router->maxwell_compat to skip all the "maxwell" queries.

Including the side-effect-free `SELECT`s or you meant just the `SET` query?  I refrained from doing this because it will complicate the state machine transition logic.

MassimilianoPinto:
Hi, please skip all the SETs and you can keep the  SELECTs.

I'm also working on the state machine transition logic for another task, so I could think about that in another time.

adaszko:
Please do not merge before I green-light this.

adaszko:
OK, tested on my end.  The ball is on your side now 🙂

MassimilianoPinto:
The code looks good.

I would only suggest that :

```
SELECT  buf = blr_make_query(router->master, "SELECT @@session.auto_increment_increment AS auto_increme ..."
static const char mysql_connector_server_variables_query[] = "SELECT @@session.auto_increment_increment AS auto_increm...
```

And
```
buf = blr_make_query(router->master, "SET sql_mode='NO_AUTO_CREATE_US ..."
static const char mysql_connector_sql_mode_query[] = "SET sql_mode='NO_AUTO_CREATE_US ..."

```
these queries could be added as defines in blr.h in order to remove duplicates.

adaszko:
> these queries could be added as defines in blr.h in order to remove duplicates.

Good idea.  Done.

markus456:

adaszko:
I hereby declare that all the code submitted by me in this pull request are subject to the 3-clause BSD license, including one or several files that are either new files or modified ones.

MassimilianoPinto:
Merge done:

https://github.com/mariadb-corporation/MaxScale/commit/825782799fd799545024dc83bd0a63f427f21c94

PR #121, opened 2017-02-14
Plugin developer manual ready for comments, version 2!

ekorh475:
Does it have everything required?


Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
+# Custom plugin development manual for MariaDB MaxScale
+
+This document explains prospective plugin developers the MariaDB MaxScale plugin API and how the core program expects the plugins to work. It will also lay out some best practices and possible pitfalls in module development. We predict that *filters* and *routers* are the module types outside developers are most likely to work on, so the APIs of these two are discussed in detail. The other module types are only glossed over in the current version of this document.
+
+This file should be included with the plugin development package.
+
+## Introduction
+
+MariaDB MaxScale is designed to be an extensible program, and as a result is something of a "swiss army-knife" of SQL-proxies. Much, if not most, of the actual processing is done by plugin modules. Plugins receive network data, process it and relay it to its destination. The MaxScale core loads plugins, manages client sessions and threads and, most importantly, offers a selection of functions for the plugins to call upon. This collection of functions is called the *Core Public Interface* or just *CPI* for short. (**or something else??**).
+
```
ekorh475: aaa

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
+# Custom plugin development manual for MariaDB MaxScale
+
+This document explains prospective plugin developers the MariaDB MaxScale plugin API and how the core program expects the plugins to work. It will also lay out some best practices and possible pitfalls in module development. We predict that *filters* and *routers* are the module types outside developers are most likely to work on, so the APIs of these two are discussed in detail. The other module types are only glossed over in the current version of this document.
```
jhnwkmn: _outside developers_ sounds a bit demeaning. This is an open source project, so we are all one big happy family.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+This file should be included with the plugin development package.
```
jhnwkmn: I suppose this line ("This file should be ...") will be removed in the final version?

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Introduction
+
+MariaDB MaxScale is designed to be an extensible program, and as a result is something of a "swiss army-knife" of SQL-proxies. Much, if not most, of the actual processing is done by plugin modules. Plugins receive network data, process it and relay it to its destination. The MaxScale core loads plugins, manages client sessions and threads and, most importantly, offers a selection of functions for the plugins to call upon. This collection of functions is called the *Core Public Interface* or just *CPI* for short. (**or something else??**).
```
jhnwkmn: _MPI_ as i MaxScale Public Interface would be nice as everything else also start with an M.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+The result of compiling a plugin should be a single shared library file. MariaDB MaxScale expects the filename to be `lib<name>.so`, where `<name>` must match the module name given in the configuration file.
+
+## Module categories
+
+This section lists all the module types and summarises their core tasks. The modules are listed in the order a client packet would typically travel through. For more information about a particular module type, see the corresponding folder in `MaxScale/Documentation/`.
+
+### Protocol
+
+Protocol modules implement I/O between clients and MaxScale, and between MaxScale and backend servers. Protocol modules read and write to socket descriptors using *raw* I/O functions provided by the CPI, and expose protocol-specific I/O functions for other modules to use. The Protocol module API is defined in `protocol.h`.
```
jhnwkmn: The protocol plugins should actually not expose general purpose functionality for other modules to use. For each protocol there ought to be a utility library exposing general purpose functionality that the protocol plugins and other modules can use.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+Currently, the only implemented database protocol is *mySQL*. Other protocols currently in use include *HTTPD* and *maxscaled*-protocols, which are used by the MaxInfo and MaxAdmin modules.
```
jhnwkmn: Why is _mySQL_ with small initial letter?

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Filter
+
+Filter modules process data from client before routing. A data buffer may travel through multiple filters before arriving in a router. For a data buffer going from a backend to the client, the router receives it first and the filters receive it in reverse order. MaxScale includes a healthly selection of filters ranging from logging, overwriting query data and caching. The filter API is defined in `filter.h`.
+
+### Router
+
+Router modules route packets from the last filter in the filter chain to backends and reply data from backends to the last filter. The routing decisions may be based on a variety of conditions; typically packet contents and backend status are the most significant factors. Router are often used for load balancing, dividing clients and even individual queries between backends. Routers use protocol functions to write to backends, making them somewhat protocol-agnostic. The router API is defined in `router.h`.
```
jhnwkmn: "Router are often used" -> "Routers are often used"

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Router
+
+Router modules route packets from the last filter in the filter chain to backends and reply data from backends to the last filter. The routing decisions may be based on a variety of conditions; typically packet contents and backend status are the most significant factors. Router are often used for load balancing, dividing clients and even individual queries between backends. Routers use protocol functions to write to backends, making them somewhat protocol-agnostic. The router API is defined in `router.h`.    
+
+### Monitor
+
+Monitor modules do not process data flowing through MariaDB MaxScale, but support the other modules (**OR JUST routers?**) in their operation by updating the status of the backend servers. Monitors are ran in their own threads to minimize interference. They periodically connect to all their assigned backends, query their status and write the results in global structs. The monitor API is defined in `monitor.h`
```
jhnwkmn: Anybody who has access to the server structure can check the server state. Filters can do that via the session and from there the service.

Monitors run in their own threads because they have to; the worker threads can only be used by entities that are capable of using asynchronous I/O on raw sockets. Since the monitors use the mysql API they must run in other threads.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+### A note on module dependencies
+
+In the ideal case modules other than the protocol modules themselves should not be protocol-speficic. This does not apply in reality, since many actions in the modules are dependent on protocol-speficic details. In the future, protocol modules may be expanded to implement a generic query parsing and information API, allowing filters and routers to be used with different SQL variants. For now, however, module writers should not focus on protocol-independence.
```
jhnwkmn: Perhaps unnecessary to point out that module writers should not focus on protocol independence. That's the reality at the moment, but already in 2.2 the situation will be different.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+`debug.h` | Debugging macros
+`modinfo.h` | Module information structure
+`server.h` | Backend server information
+`service.h` | Service definition
+`session.h` | Client session definition
+`logmanager.h` | Logging macros and functions
+
+### Module information container
+
+The module must implement the `MXS_CREATE_MODULE()`-function, which returns a pointer to a statically allocated `MXS_MODULE`-structure. This function is called by the module loader during program startup. `MXS_MODULE` (type defined in `modinfo.h`) contains function pointers to further module entrypoints, miscellaneous information about the module and the configuration parameters accepted by the module. This function must be exported without C++ name mangling, so in C++ code it should be defined `extern "C"`. Please see *RoundRobinRouter* source for a concrete example.
```
jhnwkmn: The module structure need not be statically allocated.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Module API overview
+
+This section explains some general concepts encountered when implementing a module API. Most of this section will apply to most module types. For more detailed information see the module specific section, header files or the doxygen documentation.
+
+Modules with configuration data define an *INSTANCE* object, which is created by the module code in a `createInstance`-function or equivalent. The instance creation function is called during MaxScale startup, usually when creating services. The module instance data is held by the *SERVICE*-structure (or other higher level construct) and given as a parameter when calling functions from the module in question. The instance structure should contain all non-client-specific information required by the functions of the module. The core does not know what the object contains (since it is defined by the module itself), nor will it modify the pointer or the referenced object in any way. In the interface, the instance pointer is defined as `void*`, but in actual module code the pointer should always be cast to the actual instance type.
```
jhnwkmn: Instance pointer is defined as _MXS_ROUTER_ or _MXS_FILTER_, which currently is, but need not be void*, so it is better not to say that. It is simply an opaque type that the return value should be cast to.

It's stronger than "should always be cast"; if it contains meaningful data, it must be cast.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+Descriptor control blocks (`DCB`), are generalized I/O descriptor types. DCBs store the file descriptor, state, remote address, username, session, and other data. DCBs are created whenever a new socket is created. Typically this happens when a new client connects or MaxScale connects the client session to backend servers. The module writer should use DCB handling functions provided by the CPI to manage connections instead of calling general networking libraries.
+
+Network data such as client queries and backend replies are held in a rather complicated buffer container type called `GWBUF`. Multiple GWBUFs can form a linked list with type information and properties in each GWBUF-node. Each node includes a pointer to a reference counted shared buffer (`SHARED_BUF`), which finally points to a slice of the actual data. What all this means is that multiple GWBUF-chains can share some parts of data while having some parts private. The construction is meant to minimize the need for data copying. In simple situations there is no need to form a linked list and the buffer acts similar to a handle. Plugin writers should use the CPI to manipulate GWBUFs.
```
jhnwkmn: Hmm, I'm not sure it is rather complicated as it's basically just a sharable linked list of buffers :)

If you append to a GWBUF you'll have a linked list. So, it's very easy to get one. Also, unless the filter/router module in its capabilities specify that it want contiguous buffers then the module writer must assume that it gets non-contiguous ones and act accordingly.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+#### General module management
+
+```java
+int process_init()
+void process_finish()
+int thread_init()
+void thread_finish()
+```
+
+These four functions are present in all `MXS_MODULE` structs and are not part of the API of any individual module type. `process_init` and `process_finish` are called by the module loader right after loading a module and just before unloading a module. Usually, these can be set to null in `MXS_MODULE` unless the module needs some general initializations before creating any instances. `thread_init` and `thread_finish` are thread specific equivalents.
```
jhnwkmn: Modules will actually not be unloaded. Perhaps better to say that `process_finish()` is called right before MaxScale terminates.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+```java
+INSTANCE* createInstance(SERVICE* service, char** options)
+void destroyInstance(INSTANCE* instance)
+```
+
+`createInstance` should read the `options` and initialize an "instance" object for use with `service`. Often, simply saving the configuration values to fields is enough. `destroyInstance` is called when the service using the module is deallocated. It should free any resources claimed by the instance. All sessions created by this instance should be closed before calling the destructor.
+
+```java
+void* newSession(INSTANCE* instance, MXS_SESSION* mxs_session)
```
jhnwkmn: Change `void*` into `SESSION*`.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+void closeSession(INSTANCE* instance, SESSION* session)
+void freeSession(INSTANCE* instance, SESSION* session)
+```
+
+These functions manage sessions. `newSession` should allocate a router or filter session attached to the client session represented by `mxs_session`. MaxScale will pass the returned pointer to all the API entrypoints that process user data. `closeSession` should close connections the session has opened and release any resources specific to the served client. The *SESSION* structure allocated in `newSession` should not be deallocated by `closeSession` but in `freeSession`. Usually these two are called in succession by the core.
```
jhnwkmn: close and free are always called in succession.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+
+`clientReply` processes data flowing from backend back to client. For routers, this function is often much simpler than `routeQuey`, since there is only one client to route to. Depending on the router, some packets may not be routed to the client. For example, if a client query was routed to multiple backends, MaxScale will receive multiple replies while the client only expects one. Routers should pass the reply packet to last filter in the chain (reversed order) using the macro `MXS_SESSION_ROUTE_REPLY`. Filters should call the `clientReply` of the previous filter in the chain. There is no need for filters to worry about being the first filter in the chain, as this is handled transparently by the session creation routine.
```
jhnwkmn: `routeQuey` typo.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+```java
+void handleError(INSTANCE* instance,SESSION* session, GWBUF* errmsgbuf,
+                 DCB* problem_dcb, mxs_error_action_t action, bool* succp);
+```
+
+This router-only entrypoint is called if `routeQuery` returns an error value of if an error occurs in one of the connections listened to by the session. The steps an error handler typically takes depend on the nature of the `problem_dcb` and the error encountered. If `problem_dcb` is a client socket, then the session is lost and should be closed. The error handler should not do this by itself, and just report the failure by setting `succp` to false. If `problem_dcb` is a backend socket, then the error handler should try to connect to another backend if the routing logic allows. If the error is simply failed authentication on the backend, then it is usually best to send the message directly to the client.
+
+
+```java
+uint64_t getCapabilities(void)
```
jhnwkmn: This now takes and `INSTANCE` pointer.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+void *(*startMonitor)(MXS_MONITOR *monitor, const MXS_CONFIG_PARAMETER *params);
+void (*stopMonitor)(MXS_MONITOR *monitor);
+void (*diagnostics)(DCB *, const MXS_MONITOR *);
+```
+
+Monitor modules typically run a repeated monitor routine with a used defined interval. The `MXS_MONITOR` is a standard monitor definition used for all monitors and contains a void pointer for storing module specific data. `startMonitor` should create a new thread for itself using functions in the CPI and have it regularly run a monitor loop. In the beginning of every monitor loop, the monitor should lock the `SERVER`-structures of its servers. This prevents any administrative action from interfering with the monitor during its pass.
+
+## Core public interface
+
+One could spend pages explaining the public headers in detail, what to do?
```
jhnwkmn: A link to the doxygen generated documents is fine at this point.

Documentation/Plugin-developer-manual.md
```diff
@@ -0,0 +1,230 @@
...
+After all dependencies are accounted for, the module should compile with a command along the lines of
+```
+gcc -Iinclude/ -I/usr/include/mariadb -shared -fPIC -g -o libmymodule.so mymodule.cpp
+```
+For large modules one may need to set up more complicated compilation, but that is not in the scope of this document.
+
+
+## Running, testing and profiling?
+
+Is this section needed?
```
jhnwkmn: A complete step by step procedure for building a module, installing it in MaxScale, starting MaxScale, seeing how the module logs some startup noise, connecting to MaxScale via maxadmin and via maxadmin causing the module to print "Hello, World!" would be informative.

PR #122, opened 2017-02-20
add build information to --version-full

timofeyturenko:
Extension for --version-full command. During the build process values of environmental variables from Jenkins  'source', 'value', 'cmake_flags' and BUILD_TAG are being copied to version.h and then this info available via --version-full command. If at the build time variables are not set - no additional printing.


markus456:
This has been merged.

PR #123, opened 2017-02-27
Spinlock bugfix: Add missing volatile qualifier

adaszko:
I happened to set `heartbeat=1` in `maxscale.cnf`.  This made master reconnects more frequent and increased contention on binlog router spinlock.  That, in turn, has resulted in MaxScale completely freezing (didn't even respond to maxadmin commands) around the time of a reconnect.  I investigated and tracked it down to a bug in the spinlock implementation.  The lack of `volatile` qualifier causes gcc optimizer to assume that the `while (lock->lock) {}` loop in `spinlock.c` is (roughly) equivalent to `for (;;);` (the `=>` arrow).  Here's the disassembly I got from gdb:

```
(gdb) disassemble spinlock_acquire
Dump of assembler code for function spinlock_acquire:
   0x00007fe401344e40 <+0>:	mov    $0x1,%edx
   0x00007fe401344e45 <+5>:	jmp    0x7fe401344e56 <spinlock_acquire+22>
   0x00007fe401344e47 <+7>:	nopw   0x0(%rax,%rax,1)
   0x00007fe401344e50 <+16>:	mov    (%rdi),%eax
   0x00007fe401344e52 <+18>:	test   %eax,%eax
   0x00007fe401344e54 <+20>:	jne    0x7fe401344e60 <spinlock_acquire+32>
   0x00007fe401344e56 <+22>:	mov    %edx,%eax
   0x00007fe401344e58 <+24>:	xchg   %eax,(%rdi)
   0x00007fe401344e5a <+26>:	test   %eax,%eax
   0x00007fe401344e5c <+28>:	jne    0x7fe401344e50 <spinlock_acquire+16>
   0x00007fe401344e5e <+30>:	repz retq
=> 0x00007fe401344e60 <+32>:	jmp    0x7fe401344e60 <spinlock_acquire+32>
End of assembler dump.
```

With `volatile` OTOH:

```C
typedef struct spinlock
{
    volatile int lock;         /*< Is the lock held? */
    ...
} SPINLOCK;
```

the contents of `lock->lock` are re-fetched on every loop iteration:

```
(gdb) disas spinlock_acquire
Dump of assembler code for function spinlock_acquire:
   0x00007f198461be40 <+0>:	mov    $0x1,%edx
   0x00007f198461be45 <+5>:	nopl   (%rax)
   0x00007f198461be48 <+8>:	mov    %edx,%eax
   0x00007f198461be4a <+10>:	xchg   %eax,(%rdi)
   0x00007f198461be4c <+12>:	test   %eax,%eax
   0x00007f198461be4e <+14>:	je     0x7f198461be58 <spinlock_acquire+24>
   0x00007f198461be50 <+16>:	mov    (%rdi),%eax
   0x00007f198461be52 <+18>:	test   %eax,%eax
   0x00007f198461be54 <+20>:	jne    0x7f198461be50 <spinlock_acquire+16>
   0x00007f198461be56 <+22>:	jmp    0x7f198461be48 <spinlock_acquire+8>
   0x00007f198461be58 <+24>:	repz retq
End of assembler dump.
```

This change fixed the freezing problem.

### DIFF
--- a/include/maxscale/spinlock.h
+++ b/include/maxscale/spinlock.h
@@ -42,7 +42,7 @@ MXS_BEGIN_DECLS
  */
 typedef struct spinlock
 {
-    int lock;         /*< Is the lock held? */
+    volatile int lock;         /*< Is the lock held? */
 #if SPINLOCK_PROFILE
     int spins;        /*< Number of spins on this lock */
     int maxspins;     /*< Max no of spins to acquire lock */


jhnwkmn:
Thanks. We have actually noticed the same thing and are in the process of addressing it. The reason it's become a problem now is that we have turned optimizations on.

adaszko:
Closing as the changes were made by someone else.

svoj:
If a spinlock implementation needs volatile to operate - it is most probably broken. Indeed spinlock of maxscale seem to have a bunch of pitfalls. I suggest somebody should review and fix it properly.

adaszko:
> If a spinlock implementation needs volatile to operate - it is most probably broken.

Why would that be?  Are you saying GCC intrinsics should be all that is needed?

svoj:
Yes, GCC intrinsics are enough.

jhnwkmn:
@svoj, could you be more specific regarding the pitfalls? That's very old code and apparently the `while (lock->lock)` was added in an attempt to make it more efficient. However, now with volatile in place I think the implementation is correct. But I assume that if that extra busy loop were removed and the unlocking done using `__sync_lock_release` no volatile would be needed. Is that what you are suggesting with the comment that GCC intrinsics are enough?

svoj:
Yes, mostly. I will go through spinlock code and post my suggestions there.

In this particular case volatile kind of alright, it disables some compiler optimisations and thus we get fresh value from memory.

The problems with volatile in this case are:
- it can be cached by memory controller (both store and load), delaying acquisition
- it generates a lot of bus traffic, which is not something we actually want

See also: http://stackoverflow.com/questions/7872175/c-volatile-variables-and-cache-memory

Possible solutions:
- add compiler memory barrier, which doesn't solve memory controller cache issue
- much better to add a delay/yield between test-and-set

svoj:
Apparently it is impossible to comment source code in github, only commits. So here're issues I found:
- ```atomic_add()``` - generally these days x86 assembly code for atomic operation is not needed: we removed it from server 10.3, use appropriate atomic built-ins instead; it doesn't have memory barrier argument; it have to issue "full" memory barrier, which is overkill for mutex implementation.
- calling ```atomic_add()``` twice (when mutex is busy) is evil
- ```while (lock->lock)``` already discussed above
- I'd avoid updating ```lock->waiting``` when mutex can be acquired immediately for scalability reasons (OTOH this seem to be debug code)
- int functions returning TRUE/FASLE are evil
- profiling is kind of alright unless it is used in production
- ```spinlock_release()``` should use __sync_lock_release as discussed above
- ```spinlock_stats()``` is subject of division by zero on integer overflow

svoj:
I'd also suggest to consider using C11 like API. See include/my_atomic.h in 10.3 as an example (it is still overcomplicated, but we're heading towards simplicity).

jhnwkmn:
Thanks!
> int functions returning TRUE/FASLE are evil

Oh, how I agree with that, but Rome wasn't built in a day. There is cleanup to do in other places as well :)

PR #163, opened 2018-01-26
use 'mdbci destroy' instead of 'vagrant destroy'

timofeyturenko:
mdbci destroy also removes VM direcory and config files and force removing VMs if 'vagrant destroy' fails


BUILD/mdbci/build.sh
```diff
@@ -66,9 +66,7 @@ $(<${script_dir}/templates/build.json.template)
 
 	# destroying existing box
 	if [ -d "$MDBCI_VM_PATH/${name}" ]; then
-		cd $MDBCI_VM_PATH/${name}
-		vagrant destroy -f
-		cd ${dir}
+                ${mdbci_dir}/mdbci destroy $name
```
amvasilyev: According to the test on 68th string, you check for configuration in the `$MDBCI_VM_PATH` directory, but here you relative path to the configuration. I think that absolute path is necessary here too.

BUILD/mdbci/build.sh
```diff
@@ -122,11 +118,7 @@ if [ "$try_already_running" == "yes" ] ; then
 fi
 if [[ "$do_not_destroy_vm" != "yes" && "$try_already_running" != "yes" ]] ; then
   echo "Destroying VM"
-  vagrant destroy -f
-  cd ..
-  rm -rf $name
-  rm -rf ${name}.json
-  rm -rf ${name}_netwotk_config
+  ${mdbci_dir}/mdbci destroy $name
```
amvasilyev, line 121: Same comment here. You use relative path to the configuration, not the absolute path.

Here you have also deleted the change of the working directory, so the code won't work.

The same issue of calling destroy with the relative path to the configuration can be seen in all

maxscale-system-test/mdbci/run_test_snapshot.sh
```diff
@@ -48,15 +48,15 @@ export repo_dir=$dir/repo.d/
...
 
 if [ $? != 0 ]; then
-	${script_dir}/destroy.sh
+	${mdbci_dir}/mdbci destroy $name
 	${MDBCI_VM_PATH}/scripts/clean_vms.sh $name
 
 	${script_dir}/create_config.sh
-       checkExitStatus $? "Error creating configuration" $snapshot_lock_file
-       . ${script_dir}/configure_backend.sh
+        checkExitStatus $? "Error creating configuration" $snapshot_lock_file
```
amvasilyev: Extra spaces were added, I think that it is a typo.

amvasilyev (commented):
In most cases destroy command is called using the relative path to the configuration. I think that it should be replaced with the absolute path.

BUILD/mdbci/build.sh
timofeyturenko: $MDBCI_VM_PATH is in use by default.  $name is a name of configuration, not a path. If MDBCI uses some other way to interpret this parameter it has to be fixed

BUILD/mdbci/build.sh
timofeyturenko: $name is not a path

PR #164, opened 2018-01-27
MXS-199: Support Causal Read in Read Write Splitting

ybbct:
1. get latest gtid position from ok packet which came from master;
2. add select master_wait_gtid function in front of client query;
3. pack wait function and client query into one multi statements query command
4. add two parameters:

     |name|values|
     |-----|------|
     |enable_causal_read | boolean|
     | causal_read_timeout| integer or float|

     when causal_read is enabled then retry_failed_reads is also enabled, so when wait timeout, the query will be retransfer to master
5. if not timeout, discard the result of wait function and modify the origin query's response, make the sequence number right

    

license.

### DIFF
--- a/include/maxscale/buffer.h
+++ b/include/maxscale/buffer.h
@@ -170,6 +170,10 @@ typedef struct gwbuf
     ((b)->end = bytes > ((char *)(b)->end - (char *)(b)->start) ? (b)->start : \
      (void *)((char *)(b)->end - (bytes)));
 
+#define GWBUF_LTRIM(b, bytes)\
+    ((b)->start = bytes > ((char *)(b)->end - (char *)(b)->start) ? (b)->end : \
+     (void *)((char *)(b)->start + (bytes)));
+
 #define GWBUF_TYPE(b) (b)->gwbuf_type
 /*<
  * Function prototypes for the API to maniplate the buffers
@@ -425,12 +429,14 @@ void gwbuf_hexdump(GWBUF* buffer);
 
 /**
  * Return pointer of the byte at offset from start of chained buffer
+ * Warning: It not guaranteed to point to a contiguous segment of memory,
+ * it is only safe to modify the first byte this pointer point to.
  *
  * @param buffer  one or more chained buffer
  * @param offset  Offset into the buffer
  * @return  if total buffer length is bigger than offset then return
  *      the offset byte pointer, otherwise return null
  */
-extern uint8_t *gwbuf_byte_point(GWBUF* buffer, size_t offset);
+extern uint8_t *gwbuf_byte_pointer(GWBUF* buffer, size_t offset);
 
 MXS_END_DECLS
--- a/include/maxscale/buffer.h
+++ b/include/maxscale/buffer.h
@@ -170,10 +170,6 @@ typedef struct gwbuf
     ((b)->end = bytes > ((char *)(b)->end - (char *)(b)->start) ? (b)->start : \
      (void *)((char *)(b)->end - (bytes)));
 
-#define GWBUF_LTRIM(b, bytes)\
-    ((b)->start = bytes > ((char *)(b)->end - (char *)(b)->start) ? (b)->end : \
-     (void *)((char *)(b)->start + (bytes)));
-
 #define GWBUF_TYPE(b) (b)->gwbuf_type
 /*<
  * Function prototypes for the API to maniplate the buffers
--- a/include/maxscale/buffer.h
+++ b/include/maxscale/buffer.h
@@ -423,4 +423,14 @@ extern void dprintAllBuffers(void *pdcb);
  */
 void gwbuf_hexdump(GWBUF* buffer);
 
+/**
+ * Return pointer of the byte at offset from start of chained buffer
+ *
+ * @param buffer  one or more chained buffer
+ * @param offset  Offset into the buffer
+ * @return  if total buffer length is bigger than offset then return
+ *      the offset byte pointer, otherwise return null
+ */
+extern uint8_t *gwbuf_byte_point(GWBUF* buffer, size_t offset);
+
 MXS_END_DECLS
--- a/include/maxscale/protocol/mysql.h
+++ b/include/maxscale/protocol/mysql.h
@@ -313,6 +313,13 @@ typedef enum
 
 static const mxs_mysql_cmd_t MXS_COM_UNDEFINED = (mxs_mysql_cmd_t) - 1;
 
+typedef enum
+{
+    EXPECTING_NOTHING = 0,
+    EXPECTING_WAIT_GTID_RESULT,
+    EXPECTING_REAL_RESULT
+} wait_gtid_state_t;
+
 /**
  * List of server commands, and number of response packets are stored here.
  * server_command_t is used in MySQLProtocol structure, so for each DCB there is
@@ -353,6 +360,8 @@ typedef struct
     int                    ignore_replies;               /*< How many replies should be discarded */
     GWBUF*                 stored_query;                 /*< Temporarily stored queries */
     bool                   collect_result;               /*< Collect the next result set as one buffer */
+    wait_gtid_state_t      wait_gtid_state;              /*< Determine boundray of wait gtid result and client query result */
+    uint32_t               next_seq;                     /*< Next packet sequence number */
     uint32_t               num_eof_packets;              /*< Encountered eof packet number, used for check packet type */
 #if defined(SS_DEBUG)
     skygw_chk_t            protocol_chk_tail;
--- a/include/maxscale/protocol/mysql.h
+++ b/include/maxscale/protocol/mysql.h
@@ -313,12 +313,6 @@ typedef enum
 
 static const mxs_mysql_cmd_t MXS_COM_UNDEFINED = (mxs_mysql_cmd_t) - 1;
 
-typedef enum
-{
-    EXPECTING_NOTHING = 0,
-    EXPECTING_WAIT_GTID_RESULT,
-    EXPECTING_REAL_RESULT
-} wait_gtid_state_t;
 
 /**
  * List of server commands, and number of response packets are stored here.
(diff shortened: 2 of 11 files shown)


server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
```diff
@@ -1134,7 +1135,44 @@ bool handle_got_target(RWSplit *inst, RWSplitSession *rses,
...
+    if (cmd == COM_QUERY && cr && rses->gtid_pos != "")
+    {
+        /**
+         * Pack wait function and client query into one multistatments query
+         * will save a round trip latency. Because no matter what the result of
+         * wait function is, the client query will be executed anyway. We don't
+         * want this happen, so use this triky way: encapsulate wait function
+         * as a subquery, when timeout the parent query will produce an error,
+         * then statements behind it will not be executed. For example:
+         * SELECT CASE WHEN MASTER_GTID_WAIT('232-1-1', 0.05) = 0 THEN 1 ELSE
```
markus456: I think a minor improvement to this can be made.
```
SET @maxscale_secret_variable=(SELECT CASE WHEN MASTER_GTID_WAIT('232-1-1', 10) = 0 THEN 1 ELSE (SELECT 1 FROM INFORMATION_SCHEMA.ENGINES) END);
```
Assign the return value of that SELECT into a user variable will return an OK packet instead of a result set. This means that only one packet is returned, an ERR or an OK packet.

markus456:
One problem with this is that the MariaDB server currently doesn't implement the GTID tracking that MySQL 5.7 does. [MDEV-11956](https://jira.mariadb.org/browse/MDEV-11956) tracks this issue and I've commented that it currently blocks MXS-119 for MariaDB.

svaroqui:
The server part is done , may be gtid can be track from session ?
https://jira.mariadb.org/browse/MDEV-8931

server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
```diff
@@ -1134,7 +1135,44 @@ bool handle_got_target(RWSplit *inst, RWSplitSession *rses,
     ss_dassert(target->session_command_count() == 0);
 
     mxs::Backend::response_type response = mxs::Backend::NO_RESPONSE;
+    MySQLProtocol *backend_proto = DCB_PROTOCOL(target->dcb(), MySQLProtocol);
```
markus456: The routers must not modify the protocol structures directly. If required, move either the protocol functionality into readwritesplit or some of the readwritesplit functionality into the protocol. Moving this into the protocol module would allow other modules to use it as well.

server/modules/routing/readwritesplit/readwritesplit.hh
```diff
@@ -166,8 +186,11 @@ struct Config
         connection_keepalive(config_get_integer(params, "connection_keepalive")),
         max_slave_replication_lag(config_get_integer(params, "max_slave_replication_lag")),
         rw_max_slave_conn_percent(0),
-        max_slave_connections(0)
+        max_slave_connections(0),
+        causal_read((causal_read_t)config_get_enum(params, causal_read_key, causal_read_values)),
+        causal_read_timeout(config_get_string(params, causal_read_timeout_key))
     {
+        retry_failed_reads = causal_read != CAUSAL_READ_NON_SUPPORT ? true: false;
```
markus456: Seems that retrying of reads will always be enabled when causal reads are disabled, even if they are explicitly disabled.

Another question I have is that why is the retrying of reads disabled when causal reads are enabled? Retrying should be safe as the same set of SQL commands will just be executed on another server.

markus456:
MariaDB doesn't support the [session_track_gtids](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_session_track_gtids) option and using [session_track_system_variables](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_session_track_system_variables) with `session_track_system_variables=last_gtid` does not return the session variable.

server/modules/protocol/MySQL/mysql_common.cc
```diff
@@ -1809,7 +1812,11 @@ void mxs_mysql_parse_ok_packet(GWBUF *buff, size_t packet_offset, size_t packet_
                 case SESSION_TRACK_STATE_CHANGE:
                 case SESSION_TRACK_SCHEMA:
                 case SESSION_TRACK_GTIDS:
-                    mxs_lestr_consume(&ptr, &size);
+                    mxs_leint_consume(&ptr);
```
markus456: Adding comments that explain what parts of the packet these commands consume would be helpful.

include/maxscale/buffer.h
```diff
@@ -423,4 +423,14 @@ extern void dprintAllBuffers(void *pdcb);
...
 
+/**
+ * Return pointer of the byte at offset from start of chained buffer
+ *
+ * @param buffer  one or more chained buffer
+ * @param offset  Offset into the buffer
+ * @return  if total buffer length is bigger than offset then return
+ *      the offset byte pointer, otherwise return null
+ */
+extern uint8_t *gwbuf_byte_point(GWBUF* buffer, size_t offset);
```
markus456: I think `gwbuf_data_pointer` would be a better name as it returns a pointer to the data in the GWBUF.

Also add a note that the returned pointer is not guaranteed to point to a contiguous segment of memory. In other words, it is only safe to modify the first byte of data that the return value points to.

markus456 (changes requested):

markus456:
One way to deal with this without having to modify the packet numbers is to use a compound statement:
```sql
BEGIN NOT ATOMIC
      SET @requested_gtid = '0-3000-999';
      SET @timeout = 10;
      IF (SELECT MASTER_GTID_WAIT(@requested_gtid, @timeout)) = 0 THEN
      SELECT 1;
      ELSE
        SET @errmsg=CONCAT('Slave `', @@hostname, '` is lagging behind too much. Current Gtid_Slave_Pos is `', @@gtid_slave_pos, '` when `', @requested_gtid, '` was requested.');
        SIGNAL SQLSTATE '45000' SET MYSQL_ERRNO=30001, MESSAGE_TEXT=@errmsg;
      END IF;
END
```
This should be understood correctly by most drivers. The problem is that this only works with [versions 10.1 and newer](https://mariadb.com/kb/en/library/using-compound-statements-outside-of-stored-programs/).

ybbct:
cool;

ybbct:
。。。 mysql do not support syntax BEGIN NOT ATOMIC

server/modules/protocol/MySQL/mariadbbackend/mysql_backend.c
```diff
@@ -775,6 +845,13 @@ gw_read_and_write(DCB *dcb)
 
         read_buffer = tmp;
 
+        discard_master_wait_gtid_result(&read_buffer, proto);
```
markus456: Same as with the parameters, this code can be moved inside readwritesplit. I think one place to put this would be in the `clientReply` function in `readwritesplit.cc`  around line 1180, before the call to `reply_is_complete`. This is the point where the readwritesplit does some processing on the result set (which really should be in the protocol module) and checks if the resultset is complete.

include/maxscale/protocol/mysql.h
```diff
@@ -353,6 +360,8 @@ typedef struct
     int                    ignore_replies;               /*< How many replies should be discarded */
     GWBUF*                 stored_query;                 /*< Temporarily stored queries */
     bool                   collect_result;               /*< Collect the next result set as one buffer */
+    wait_gtid_state_t      wait_gtid_state;              /*< Determine boundray of wait gtid result and client query result */
```
markus456: These can be moved into readwritesplit as it is the only module that uses it. The GTID tracking is useful for other modules as well so it is good to keep in the protocol module.

markus456 (commented):
Since MariaDB doesn't support GTID tracking but MySQL 5.7 does, I think this is a good starting point for MXS-199.

I would say that the code in the protocol modules should be moved into readwritesplit. This way the two modules don't depend on each other. After that, I would say this is OK to merge.

server/modules/routing/readwritesplit/readwritesplit.cc
```diff
@@ -230,6 +230,18 @@ static bool rwsplit_process_router_options(Config& config,
                     success = false;
                 }
             }
+            else if (strcmp(options[i], "enable_causal_read") == 0)
```
markus456: The options can probably be removed, they have been deprecated in MaxScale 2.2 and we will most likely remove all router options in 2.3.

include/maxscale/buffer.h
```diff
@@ -170,6 +170,10 @@ typedef struct gwbuf
     ((b)->end = bytes > ((char *)(b)->end - (char *)(b)->start) ? (b)->start : \
      (void *)((char *)(b)->end - (bytes)));
 
+#define GWBUF_LTRIM(b, bytes)\
```
markus456: This is not needed, `gwbuf_consume` already does it.

server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
```diff
@@ -1107,6 +1108,60 @@ static inline bool is_large_query(GWBUF* buf)
...
+
+    /**
+     * Trim origin to sql, Append origin buffer to the prefix buffer
+     **/
+    uint8_t header[MYSQL_HEADER_LEN];
+    gwbuf_copy_data(origin, 0, MYSQL_HEADER_LEN, header);
+    /* Command length = 1 */
+    size_t origin_sql_len = gw_mysql_get_byte3(header) - 1;
+    /* Trim mysql header and command */
+    GWBUF_LTRIM(origin, MYSQL_HEADER_LEN + 1);
```
markus456: This can be replaced with `gwbuf_consume` which also correctly handles headers split between two buffers.

markus456 (changes requested):
Starting to look good. After review comments are addressed, I think this is OK to merge.

markus456 (approved):

server/modules/routing/readwritesplit/readwritesplit.cc
```diff
@@ -1142,6 +1143,62 @@ static void log_unexpected_response(DCB* dcb, GWBUF* buffer)
...
+/**
+ * @bref discard the result of wait gtid statment, the result will be an error
+ * packet or an error packet.
+ * @param buffer origin reply buffer
+ * @param proto  MySQLProtocol
+ * @return reset buffer
+ */
+GWBUF *discard_master_wait_gtid_result(GWBUF *buffer, RWSplitSession *rses)
+{
+    uint8_t header_and_command[MYSQL_HEADER_LEN+1];
```
jhnwkmn: Space on both sides of a binary operator. Please correct in all places.

jhnwkmn (approved):

PR #166, opened 2018-02-07
Mxs 1140 keepalived

timofeyturenko:
* first implementation of tests with two Maxscales and keepalived added
* if there are several Maxscale machines and there are template files with numeric extentions (template.000,, template.001) a separate maxscale.cnf template will be installed to every Maxscale machine
* set_env.sh sets also start/stop_vm_command for every node (using vagrant suspend/resume)

TODO: install keepalived with Chef recipe instead of using ssh and yum from test code (same to be done for kerberos test)


maxscale-system-test/keepalived.cpp
```diff
@@ -0,0 +1,119 @@
...
+        Test->maxscales->copy_to_node(i, src.c_str(), Test->maxscales->access_homedir[i]);
+        Test->maxscales->ssh_node(i, cp_cmd.c_str(), true);
+        Test->maxscales->ssh_node_f(i, true, "sed -i \"s/###virtual_ip###/%s/\" /etc/keepalived/keepalived.conf", virtual_ip);
+        Test->maxscales->ssh_node(i, "sudo service keepalived restart", true);
+    }
+
+    print_version_string(Test);
+
+    Test->tprintf("Stop Maxscale 000 and waiting\n");
+    system(Test->maxscales->stop_vm_command[0]);
```
markus456: This could also test that failover happens even when just MaxScale is stopped.

markus456 (approved):

markus456 (approved):

PR #174, opened 2018-05-31
MXS-1891: dealloc named prepare should route to all

ybbct:
license.

Because prepare statement route to all, so drop/dealloc prepare should route to all too;


jhnwkmn (commented):
In principle this looks ok. However, the prepare type information should also be removed from the PS manager.

jhnwkmn:
I think this looks good but I'll let Markus also take a look before merging.

server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
```diff
@@ -345,6 +345,11 @@ bool route_session_write(RWSplitSession *rses, GWBUF *querybuf,
         rses->ps_manager.store(querybuf, id);
     }
 
+    if (qc_query_is_type(type, QUERY_TYPE_DEALLOC_PREPARE))
```
markus456: This can be combined into the previous `if` statement, the two are mutually exclusive.
```
     if (qc_query_is_type(type, QUERY_TYPE_PREPARE_NAMED_STMT) ||
         qc_query_is_type(type, QUERY_TYPE_PREPARE_STMT))
     ...
   else  if (qc_query_is_type(type, QUERY_TYPE_DEALLOC_PREPARE))
```

markus456 (approved):
Looks good.

PR #180, opened 2018-08-24
MXS-2042: fix multistatments hang

ybbct:
license.

If enable multi statements,  and all result sets returned in one packet, then it will hang;

For example
```
select name from t1 where id = 1; select name from t1 where id = 3; select name from t1 where id = 2

```

Because function `modutil_count_signal_packets` will return more than 2;


markus456:
If you have example SQL that causes this, can you open a new bug report on Jira? This does look like it solves a bug.

We can create a test case once you've filed the bug report for this.

markus456 (approved):

jhnwkmn (approved):

PR #181, opened 2018-08-29
Mxs 1796 twomaxscale tests

timofeyturenko:
license.

### DIFF
--- a/maxscale-system-test/mdbci/templates/default.json.template
+++ b/maxscale-system-test/mdbci/templates/default.json.template
@@ -104,7 +104,19 @@
         }
   },
 
-  "maxscale" :
+  "maxscale_000" :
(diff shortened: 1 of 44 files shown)


maxscale-system-test/mdbci/templates/default.json.template
```diff
@@ -104,7 +104,19 @@
...
+        "product" : {
+                 "name" : "maxscale_ci",
+                "version" : "${target}"
+        }
+
+  },
+
+  "maxscale_001" :
   {
         "hostname" : "maxscale",
```
markus456: Both have the same hostname, might cause problems. Changing it to `maxscale2` would be OK.

markus456 (approved):

PR #183, opened 2018-10-29
Update galeramon.c

zloma:
Support wsrep_sst_method in Galera  ("xtrabackup-v2") for "available_when_donor" maxscale option

license.


markus456 (approved):

jhnwkmn (changes requested):
Otherwise ok, but please reformat so that the maximum line length is <= 110 chars.

jhnwkmn (approved):
That's a bit peculiar, I was expecting something like this:

if (row[1] && (strcmp(row[1], "xtrabackup") == 0 
               || strcmp(row[1], "mariabackup") == 0 || strcmp(row[1], "xtrabackup-v2") == 0))

but it does not matter. We'll fix that.

PR #185, opened 2018-12-16
try to Fix MXS-2206

Alexander198961:
license.

There is link for issue https://jira.mariadb.org/browse/MXS-2206

### DIFF
--- a/BUILD/install_build_deps.sh
+++ b/BUILD/install_build_deps.sh
@@ -86,29 +86,9 @@ if [ $? != 0 ] ; then
     wget -q https://cmake.org/files/v3.7/cmake-3.7.1-Linux-x86_64.tar.gz --no-check-certificate
 fi
 sudo tar xzf cmake-3.7.1-Linux-x86_64.tar.gz -C /usr/ --strip-components=1
-check_cmake_version_is_greater_than_default()
-{
-  compared_major_version=3
-  compared_minor_version=7
-  compared_patch_version=1
-  IFS='.' declare -a 'arr=($1)'
-  major_version=${arr[0]}
-  minor_version=${arr[1]}
-  patch_version=${arr[2]}
-  unset IFS
- if [[ $major_version -ne $compared_major_version ]] ; then
-    [[ $major_version -gt $compared_major_version ]] && return 0 || return 1
- fi
- if [[ $minor_version -ne $compared_minor_version ]] ; then
-    [[ $minor_version -gt $compared_minor_version ]] && return 0 || return 1
- fi
- [[ $patch_version -ge $compared_patch_version ]] && return 0 || return 1
-
-}
 
 cmake_version=`cmake --version | grep "cmake version" | awk '{ print $3 }'`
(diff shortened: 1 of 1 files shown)


BUILD/install_build_deps.sh
```diff
@@ -87,9 +86,29 @@ if [ $? != 0 ] ; then
...
+    [[ $minor_version -gt $compared_minor_version ]] && return 0 || return 1
+ fi
+ [[ $patch_version -ge $compared_patch_version ]] && return 0 || return 1
+
+}
 
 cmake_version=`cmake --version | grep "cmake version" | awk '{ print $3 }'`
-if [ "$cmake_version" \< "3.7.1" ] ; then
+check_cmake_version_is_greater_than_default $cmake_version
+if [ $? -gt 0 ] ; then
```
markus456: We can use the `sort` utility here as all platforms that MaxScale supports also support the version sorting flag `-V`.
```suggestion
 if [ "`echo -e "3.7.1\n$cmake_version"|sort -V|head -n 1`" == "3.7.1" ] ; then
```

markus456 (commented):
Looks OK but the code could be made quite a lot simpler by using the `sort` utility. It's available on all of the supported platforms so there's no problem in using it.

BUILD/install_build_deps.sh
Alexander198961: I will test this to see if it works for me

timofeyturenko: > We can use the `sort` utility here as all platforms that MaxScale supports also support the version sorting flag `-V`.
> 
> Suggested change Beta 
> [Give us feedback](https://www.research.net/r/SuggestedChanges)
> if [ $? -gt 0 ] ; then
>  if [ "`echo -e "3.7.1\n$cmake_version"|sort -V|head -n 1`" == "3.7.1" ] ; then

one thing: we need to be sure that 'sort' is installed on the build machine. It belongs to "coreutils" package. Most probably it is always installed as dependency, but let's check it

Alexander198961: There is my test  for example if we have version 3.6.1 then
echo -e "3.7.1\n3.6.1" |sort -V|head -n 1  returns for me 3.6.1 and then we compare with "3.7.1"  "3.7.1"  so it should be != ""3.7.1"" then download (the order is 3.6.1 then 3.7.1)

Alexander198961: @timofeyturenko coreutils  package include  commands like cat && mkdir which we have already used in the script so maybe we can skip this check  )  @markus456 I added your suggestion just replaced    "=="  with "!="     because of if we have cmake which has version 3.6.1 and then run echo -e "3.6.1\n3.7.1" | sort -V 
3.6.1
3.7.1
 then take head command   and we got 3.6.1 In this case we should download 3.7.1 (3.6.1 != 3.7.1)

markus456: Yes, that does seem correct. I think I was testing it on the command line and I got them mixed up the other way around.

markus456 (approved):

PR #191, opened 2019-03-26
Fix wsrep_local_state when donor

woopstar:
According to the docs about node state changes at http://galeracluster.com/documentation-webpages/nodestates.html#node-state-changes the `wsrep_local_state` would be `5` when it is in donor mode and not `2`

license.

### DIFF
--- a/server/modules/monitor/galeramon/galeramon.cc
+++ b/server/modules/monitor/galeramon/galeramon.cc
@@ -194,7 +194,7 @@ void GaleraMonitor::update_server_status(MXS_MONITORED_SERVER* monitored_server)
                     info.joined = 1;
                 }
                 /* Check if the node is a donor and is using xtrabackup, in this case it can stay alive */
-                else if (strcmp(row[1], "2") == 0 && m_availableWhenDonor == 1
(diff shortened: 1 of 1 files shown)


markus456 (approved):

woopstar:
Hold on this. According to https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-status-index.html the state is actually 2 when either joining or as donor.

But if you look at the MariaDB JDBC connector at https://mariadb.com/kb/en/library/about-mariadb-connector-j/ you’ll see there is an option to check the allowed Galera states. The option is `galeraAllowedState` and the example says 4,5 and refers to same page as I did.

So this is a bit confusing what the local state actually is.

markus456:
Based on the documentation it does look like it should be 5 but I when I think about it I do remember seeing it work with 2. I think we'll have to investigate this before merging. I'll ask around and see whether some people I know have a better idea of this.

markus456:
Since it seems to be working like it is now, I'll close this pull request. Perhaps the documentation for Galera is somehow mistaken on which states actually appear.

PR #194, opened 2019-05-07
add missing grant in MaxScale-Tutorial.md

ktugan:
license.


markus456 (approved):

markus456:
Strictly speaking it's not required by the MaxScale user but the user used for `mariadbmon`. Although not necessary, it is relatively common to combine both accounts into the same one.

ktugan:
Good to know. I can also update the readme to reflect this if interested.

markus456:
I think it would be better to have a separate paragraph that describes the grants that the monitor user requires. To make the documentation easier to maintain, it should also have a link to the documentation of the most frequently used monitors (mariadbmon and galeramon).

markus456:
We'll close this pull request as the tutorials will be formatted a bit differently in the future.

PR #195, opened 2019-06-20
MXS-2521:Route subseqenct COM_STMT_EXECUTE to the same server which f…

ybbct:
In the scenario as below

```
 mysql_stmt_prepare
 mysql_stmt_bind_param
 mysql_stmt_execute
 mysql_stmt_execute
 mysql_stmt_execute
 ...
```

First mysql_stmt_execute will produce packet with params meta info, but the rest mysql_stmt_executes do not have the same info,  if first mysql_stmt_execute go to server1, and second go to server2, will produce different result; because server2 do not have these meta info will produce unexpected results;

image

I created a simple test case to reproduce this problem, enable general_log, we can see that 'real query' execute on master and slave are different:
```
 master general_log: select c1 from test_sub_execute where c1= 8.0
 slave general_log : select c1 from test_sub_execute where c1= NULL
```

this PR will record n_params to parse flag new-params-bound-flag, 
if (new-params-bound-flag = 0 && n_params > 0) route COM_STMT_EXECUTE to the server which last COM_STMT_EXECUTE  executed on 

ps: test case https://jira.mariadb.org/secure/attachment/48236/48236_test-mxs-2521.c

license.

### DIFF
--- a/server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
+++ b/server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
@@ -1002,7 +1002,7 @@ SRWBackend handle_hinted_target(RWSplitSession *rses, GWBUF *querybuf,
 /**
  * @brief Determine whether this stmt  is subsequent COM_STMT_EXECUTE
  *
- * @param cmd     command type 
+ * @param cmd     command type
  * @param query   GWBUF including the query
  * @param info    Route info
  *
@@ -1021,9 +1021,9 @@ bool is_sub_stmt_exec(uint8_t cmd, const GWBUF *query, uint16_t n_params)
     /*need n_params to parse new_params_bound_flag(alias send type to server)*/
     int new_params_bound_flag_offset = MYSQL_HEADER_LEN + 10 + (n_params + 7) / 8;
     ss_dassert((int)gwbuf_length(query) >= new_params_bound_flag_offset);
-    uint8_t data[new_params_bound_flag_offset];
-    gwbuf_copy_data(query, 0, new_params_bound_flag_offset, data);
-    if (data[new_params_bound_flag_offset])
+    uint8_t flag;
+    gwbuf_copy_data(query, new_params_bound_flag_offset, 1, &flag);
+    if (flag)
     {
         rval = false;
     }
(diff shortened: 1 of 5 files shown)


server/modules/routing/readwritesplit/rwsplit_session_cmd.cc
```diff
@@ -122,7 +122,7 @@ void process_sescmd_response(RWSplitSession* rses, SRWBackend& backend,
                     {
                         /** Map the returned response to the internal ID */
                         MXS_INFO("PS ID %u maps to internal ID %lu", resp.id, id);
-                        rses->ps_handles[resp.id] = id;
+                        rses->ps_handles[resp.id] = (id << 16) + resp.parameters;
```
markus456: Left shift should be 32 bits since the stored value is 64 bits and the ID is 32.

server/modules/routing/readwritesplit/rwsplitsession.cc
```diff
@@ -120,7 +120,8 @@ uint32_t get_internal_ps_id(RWSplitSession* rses, GWBUF* buffer)
 
     if (it != rses->ps_handles.end())
     {
-        rval = it->second;
+        rval = it->second >> 16;
```
markus456: Same here, shift 32 bits.

markus456 (commented):
This needs to be rewritten for the 2.3 branch as the code is quite different and can be implemented in a cleaner manner there.

server/modules/routing/readwritesplit/rwsplit_route_stmt.cc
```diff
@@ -999,45 +999,84 @@ SRWBackend handle_hinted_target(RWSplitSession *rses, GWBUF *querybuf,
...
+    }
+
+    bool rval = true;
+
+    /*https://mariadb.com/kb/en/library/com_stmt_execute/*/
+    /*need n_params to parse new_params_bound_flag(alias send type to server)*/
+    int new_params_bound_flag_offset = MYSQL_HEADER_LEN + 10 + (n_params + 7) / 8;
+    ss_dassert((int)gwbuf_length(query) >= new_params_bound_flag_offset);
+    uint8_t data[new_params_bound_flag_offset];
+    gwbuf_copy_data(query, 0, new_params_bound_flag_offset, data);
```
markus456: Can be optimized from:
```c++
    uint8_t data[new_params_bound_flag_offset];
    gwbuf_copy_data(query, 0, new_params_bound_flag_offset, data);
```
to:
```c++
    uint8_t data;
    gwbuf_copy_data(query, new_params_bound_flag_offset, 1, &data);
```
That way only one byte of data is copied.

server/modules/routing/readwritesplit/rwsplitsession.hh
```diff
@@ -202,7 +203,8 @@ static inline const char* rstostr(reply_state_t state)
  *
  * @param rses   Router client session
  * @param buffer Buffer containing a binary protocol statement other than COM_STMT_PREPARE
+ * @param n_params statement parmas number
```
markus456:
```suggestion
 * @param n_params Statement params count
```

markus456 (approved):
Looks OK for 2.2. For 2.3, most of the work can be done inside the query classifier. With this code, it should be quite easy to merge into 2.3 and do the required modifications on our side.

PR #198, opened 2019-08-09
MXS-2628: load balance use lru when score is same

ybbct:
license.

### DIFF
--- a/include/maxscale/service.h
+++ b/include/maxscale/service.h
@@ -65,7 +65,6 @@ typedef struct server_ref_t
     double               server_weight; /**< Weight in the range [0..1]. 0 is worst, and a special case. */
     int                  connections;   /**< Number of connections created through this reference */
     bool                 active;        /**< Whether this reference is valid and in use*/
-    uint64_t              lru_clock;
 } SERVER_REF;
 
 /** Macro to check whether a SERVER_REF is active */
@@ -280,8 +279,5 @@ typedef enum service_version_which_t
  * @return The version of the service.
  */
 uint64_t service_get_version(const SERVICE* service, service_version_which_t which);
-
-// return lru clock
-uint64_t mxs_lru_clock();
 
 MXS_END_DECLS
--- a/include/maxscale/service.h
+++ b/include/maxscale/service.h
@@ -65,6 +65,7 @@ typedef struct server_ref_t
     double               server_weight; /**< Weight in the range [0..1]. 0 is worst, and a special case. */
     int                  connections;   /**< Number of connections created through this reference */
     bool                 active;        /**< Whether this reference is valid and in use*/
+    uint64_t              lru_clock;
 } SERVER_REF;
 
 /** Macro to check whether a SERVER_REF is active */
@@ -279,5 +280,8 @@ typedef enum service_version_which_t
  * @return The version of the service.
  */
 uint64_t service_get_version(const SERVICE* service, service_version_which_t which);
+
+// return lru clock
+uint64_t mxs_lru_clock();
 
 MXS_END_DECLS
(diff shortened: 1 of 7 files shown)


server/core/service.cc
```diff
@@ -2425,3 +2427,8 @@ void Service::update_basic_parameter(const std::string& key, const std::string&
         retain_last_statements = std::stoi(value);
     }
 }
+
+uint64_t mxs_lru_clock()
```
markus456: This could probably be renamed from `mxs_lru_clock` to `mxs_logical_clock_tick` to tell the reader that it modifies the value.

include/maxscale/service.h
```diff
@@ -65,6 +65,7 @@ typedef struct server_ref_t
     double               server_weight; /**< Weight in the range [0..1]. 0 is worst, and a special case. */
     int                  connections;   /**< Number of connections created through this reference */
     bool                 active;        /**< Whether this reference is valid and in use*/
+    uint64_t              lru_clock;
```
markus456: This could be declared inside the function as a static variable to reduce the scope.

server/core/backend.cc
```diff
@@ -223,6 +223,7 @@ bool Backend::connect(MXS_SESSION* session, SessionCommandList* sescmd)
 bool Backend::write(GWBUF* buffer, response_type type)
 {
     mxb_assert(in_use());
+    m_backend->lru_clock = mxs_lru_clock();
```
markus456: The way Backend uses it is different from how readconnroute uses it. Backend sets the clock on each write whereas readconnroute sets it on each connection. Since these are shared between different routers, they should behave similarly.

server/modules/routing/readconnroute/readconnroute.cc
```diff
@@ -367,11 +367,20 @@ static MXS_ROUTER_SESSION* newSession(MXS_ROUTER* instance, MXS_SESSION* session
...
-                /* ref has a better score. */
-                candidate = ref;
+                uint64_t score1 = (ref->connections + 1) / ref->server_weight;
+                uint64_t score2 = (candidate->connections + 1) / candidate->server_weight;
+                if (score1 < score2)
+                {
+                    /* ref has a better score. */
+                    candidate = ref;
+                }
+                else if (score1 == score2 && ref->lru_clock < candidate->lru_clock)
```
markus456: How often do you see this happening? If it is rare, a random probability of picking one of the servers might be simpler and would remove the need for the logical clock. It wouldn't be a completely balanced probability distribution but it should at least resolve the "stalemate problem" that happens when the score is the same.

ybbct: Last week we saw an online instance (with 3k qps, readwritesplit router),  almost 87% queries go to same node, so it is not rare;

ybbct: This logic we had added to 2.2 in May，and it perform well, before that we almost receive ticket about this problem every week;

ybbct:
Sorry, we are not using readconn router , maybe readconn is ok

server/modules/routing/readconnroute/readconnroute.cc
markus456: If 87% of the queries go to only one node that would suggest that a non-default load balancing algorithm was in use. The default (LEAST_CURRENT_OPERATIONS) should spread out queries evenly across all connected servers.

In 2.4 we changed the code to always use the number of connections when creating new connections and the load balancing algorithm only for queries. Have you seen this with that version?

markus456 (approved):
I think this might fix the same thing that the change in 2.4 does with the connection selection.

server/modules/routing/readconnroute/readconnroute.cc
ybbct: No, most of our instances are 2.2, few are 2.3, no 2.4

PR #204, opened 2019-10-30
bugfix: calculate wrong weight when add server to maxscale on rumtime

lishubing:
license.

-----

**This MR fix two issues:**
1. Each time add server to service on runtime, trigger recalculate server weight. If not so, the value of `sref->server_weight` will be the default value 1.0, which should be calculated by weight parameter.

2. When a `server_ref` object is removed, instead of really remove from the link list, MaxScale simply mark them to inactive. So in calculating weight, we should skip those inactive `server_ref`.

### DIFF
--- a/server/core/config_runtime.cc
+++ b/server/core/config_runtime.cc
@@ -175,6 +175,7 @@ bool runtime_link_server(Server* server, const char* target)
         {
             if (serviceAddBackend(service, server))
             {
+                service_update_weights();
                 service_serialize(service);
                 rval = true;
             }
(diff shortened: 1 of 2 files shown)


server/core/config_runtime.cc
```diff
@@ -175,6 +175,7 @@ bool runtime_link_server(Server* server, const char* target)
         {
             if (serviceAddBackend(service, server))
```
markus456, line 176: The call to `service_update_weights` should be done inside `serviceAddBackend` and `serviceRemoveBackend`.

PR #214, opened 2020-12-02
MXS-3238

DatabasKarlsson:
Fixes to specify accepted ciphers for the REST API as per MXS-3238

### DIFF
--- a/include/maxscale/config.hh
+++ b/include/maxscale/config.hh
@@ -137,6 +137,7 @@ public:
     std::string   admin_pam_rw_service;         /**< PAM service for read-write users */
     std::string   admin_pam_ro_service;         /**< PAM service for read-only users */
     std::string   admin_ssl_key;                /**< Admin SSL key */
+    std::string   admin_ssl_cipher;             /**< Admin allowed SSL ciphers */
     std::string   admin_ssl_cert;               /**< Admin SSL cert */
     std::string   admin_ssl_ca_cert;            /**< Admin SSL CA cert */
     std::string   local_address;                /**< Local address to use when connecting */
@@ -218,6 +219,7 @@ private:
     static config::ParamString              s_admin_pam_rw_service;
     static config::ParamString              s_admin_pam_ro_service;
     static config::ParamString              s_admin_ssl_key;
+    static config::ParamString              s_admin_ssl_cipher;
     static config::ParamString              s_admin_ssl_cert;
     static config::ParamString              s_admin_ssl_ca_cert;
     static config::ParamString              s_local_address;
--- a/server/core/admin.cc
+++ b/server/core/admin.cc
@@ -284,19 +284,33 @@ static bool load_ssl_certificates()
 
     // Set cipher appropriately, including some shorcuts.
     if (cipher.empty())
+    {
         this_unit.ssl_cipher = "NORMAL";
+    }
     else if (cipher == "SSLv3")
+    {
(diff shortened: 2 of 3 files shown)


server/core/admin.cc
```diff
@@ -279,6 +280,24 @@ static bool load_ssl_certificates()
     const auto& cert = config.admin_ssl_cert;
     const auto& ca = config.admin_ssl_ca_cert;
 
+    const auto& cipher = config.admin_ssl_cipher;
+
+    // Set cipher appropriately, including some shorcuts.
+    if (cipher.empty())
```
jhnwkmn, line 286: We follow the Allman style (yes, always), so please add braces everywhere.

server/core/admin.cc
```diff
@@ -938,6 +971,7 @@ bool mxs_admin_init()
                                             MHD_OPTION_SOCK_ADDR, &addr,
                                             !this_unit.using_ssl ? MHD_OPTION_END :
                                             MHD_OPTION_HTTPS_MEM_KEY, this_unit.ssl_key.c_str(),
+                                            MHD_OPTION_HTTPS_PRIORITIES, this_unit.ssl_cipher.c_str(),
```
markus456, line 974: For the sake of consistency, I'd say add this after `MHD_OPTION_HTTPS_MEM_CERT`. That way the certificate arguments follow each other.

server/core/admin.cc
```diff
@@ -279,6 +280,38 @@ static bool load_ssl_certificates()
     const auto& cert = config.admin_ssl_cert;
     const auto& ca = config.admin_ssl_ca_cert;
 
+    const auto& cipher = config.admin_ssl_cipher;
+
+    // Set cipher appropriately, including some shorcuts.
+    if (cipher.empty())
```
markus456, line 286: A separate function that does the translation into GnuTLS values would make this cleaner. When the value of `config.admin_ssl_cipher` is changed into an enumeration, this becomes quite simpler to do.

In addition, SSLV3 is not supported by MaxScale and should not be supported by the REST API.

include/maxscale/config.hh
```diff
@@ -137,6 +137,7 @@ public:
     std::string   admin_pam_rw_service;         /**< PAM service for read-write users */
     std::string   admin_pam_ro_service;         /**< PAM service for read-only users */
     std::string   admin_ssl_key;                /**< Admin SSL key */
+    std::string   admin_ssl_cipher;             /**< Admin allowed SSL ciphers */
```
markus456, line 140: Since the original issue is not with TLS cipher configuration and the code behaves as if this was similar to the `ssl_version` parameter of listeners and servers, this should be named `admin_ssl_version` and changed into `mxb::ssl_version::Version`.

Having one variable do both is convenient from the implementation point of view but it's inconsistent from the user's point of view. If one types the name of the TLS version wrong, you'll most likely get a less than obvious error about the REST API not being able to start up.

include/maxscale/config.hh
```diff
@@ -218,6 +219,7 @@ private:
     static config::ParamString              s_admin_pam_rw_service;
     static config::ParamString              s_admin_pam_ro_service;
     static config::ParamString              s_admin_ssl_key;
+    static config::ParamString              s_admin_ssl_cipher;
```
markus456, line 222: Change this to `config::ParamEnum<mxb::ssl_version::Version>` so that unknown values are rejected.

PR #216, opened 2020-12-07
MXS-3238 and also fixed minor bug when starting libhttpd where ssl_ce…

DatabasKarlsson:
…rt was passed instead of ssl_ca

license.

### DIFF
--- a/include/maxscale/config.hh
+++ b/include/maxscale/config.hh
@@ -138,6 +138,7 @@ public:
     std::string   admin_pam_ro_service;         /**< PAM service for read-only users */
     std::string   admin_ssl_key;                /**< Admin SSL key */
     std::string   admin_ssl_cert;               /**< Admin SSL cert */
+    admin_ssl_version_t    admin_ssl_version;   /**< Admin allowed SSL versions */
     std::string   admin_ssl_ca_cert;            /**< Admin SSL CA cert */
     std::string   local_address;                /**< Local address to use when connecting */
     bool          load_persisted_configs;       /**< Load persisted configuration files on startup */
@@ -219,6 +220,7 @@ private:
     static config::ParamString              s_admin_pam_ro_service;
     static config::ParamString              s_admin_ssl_key;
     static config::ParamString              s_admin_ssl_cert;
+    static config::ParamEnum<admin_ssl_version_t>              s_admin_ssl_version;
     static config::ParamString              s_admin_ssl_ca_cert;
     static config::ParamString              s_local_address;
     static config::ParamBool                s_load_persisted_configs;
--- a/include/maxscale/config.hh
+++ b/include/maxscale/config.hh
@@ -15,6 +15,7 @@
 #include <maxscale/ccdefs.hh>
 #include <maxscale/config2.hh>
 #include <maxscale/session.hh>
+#include <maxbase/ssl.hh>
 
 namespace maxscale
 {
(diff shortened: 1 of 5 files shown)


server/core/admin.cc
```diff
@@ -939,8 +970,9 @@ bool mxs_admin_init()
                                             !this_unit.using_ssl ? MHD_OPTION_END :
                                             MHD_OPTION_HTTPS_MEM_KEY, this_unit.ssl_key.c_str(),
                                             MHD_OPTION_HTTPS_MEM_CERT, this_unit.ssl_cert.c_str(),
+                                            MHD_OPTION_HTTPS_PRIORITIES, this_unit.ssl_version.c_str(),
                                             this_unit.ssl_ca.empty() ? MHD_OPTION_END :
-                                            MHD_OPTION_HTTPS_MEM_TRUST, this_unit.ssl_cert.c_str(),
+                                            MHD_OPTION_HTTPS_MEM_TRUST, this_unit.ssl_ca.c_str(),
```
markus456, line 975: Please don't include this change in this commit, it's not related to the issue.

It seems to be a regression of [MXS-2956](https://jira.mariadb.org/browse/MXS-2956). If you want, you can submit a separate pull request for it and we'll merge it right away.

server/core/config.cc
```diff
@@ -461,6 +462,18 @@ config::ParamString Config::s_admin_ssl_key(
     "Admin SSL key",
     "");
 
+config::ParamEnum<admin_ssl_version_t> Config::s_admin_ssl_version(
+    &Config::s_specification,
+    CN_ADMIN_SSL_VERSION,
+    "Admin allowed SSL version",
+    {
+        {ADMIN_SSL_VERSION_TLS10, "TLSv1.0"},
```
markus456: Use the same values as defined here: https://mariadb.com/kb/en/mariadb-maxscale-25-mariadb-maxscale-configuration-guide/#ssl_version

In addition, the default value should be in the configuration parameter enum as it should be possible to explicitly define the default value. The default can be added as the `MAX` value as it supports the highest available TLS version that both the client and the server support.

include/maxscale/config.hh
```diff
@@ -138,6 +138,7 @@ public:
     std::string   admin_pam_ro_service;         /**< PAM service for read-only users */
     std::string   admin_ssl_key;                /**< Admin SSL key */
     std::string   admin_ssl_cert;               /**< Admin SSL cert */
+    admin_ssl_version_t    admin_ssl_version;   /**< Admin allowed SSL versions */
```
markus456: Align the variable names at the same level. In addition, add this as the last TLS related variable (after `admin_ssl_ca_cert`) to keep the certificate related parameters grouped together.

include/maxscale/config.hh
```diff
@@ -219,6 +220,7 @@ private:
     static config::ParamString              s_admin_pam_ro_service;
     static config::ParamString              s_admin_ssl_key;
     static config::ParamString              s_admin_ssl_cert;
+    static config::ParamEnum<admin_ssl_version_t>              s_admin_ssl_version;
```
markus456: Align the variable names.

include/maxscale/config_common.hh
```diff
@@ -47,6 +47,16 @@ using DiskSpaceLimits = std::unordered_map<std::string, int32_t>;
 #define DEFAULT_ADMIN_HTTP_PORT 8989
 #define DEFAULT_ADMIN_HOST      "127.0.0.1"
 
+/** SSL versions for the REST API */
+enum admin_ssl_version_t
```
markus456: Use the `maxbase::ssl_version::Version` enum from `<maxbase/ssl.hh>`, this way all the TLS version related code uses the same enumeration values.

server/core/admin.cc
```diff
@@ -271,6 +272,35 @@ std::string get_filename(const HttpRequest& request)
     return path;
 }
 
+static const char *get_ssl_version(const enum admin_ssl_version_t ssl_version)
+{
+    const char *ver;
+
+    if (ssl_version == ADMIN_SSL_VERSION_NORMAL)
```
markus456: Use a `switch` statement and handle all the known cases separately. This will prevent this code from compiling when new enumeration values are added.

server/core/admin.cc
```diff
@@ -271,6 +272,35 @@ std::string get_filename(const HttpRequest& request)
...
+    else if (ssl_version == ADMIN_SSL_VERSION_TLS12)
+    {
+        ver = "NORMAL:-VERS-SSL3.0:-VERS-TLS1.0:-VERS-TLS1.1";
+    }
+    else
+    {
+        ver = "NORMAL:-VERS-SSL3.0:-VERS-TLS1.0:-VERS-TLS1.1:-VERS-TLS1.2";
+    }
+
+    MXS_NOTICE("SSL PRIORITY for REST API set to %s", ver);
```
markus456: Don't log this, it's not useful for the end user. For developers it's probably easier to manually verify by attaching a debugger.

markus456 (commented):
Please let us know if this will replace your original pull request for MXS-3238. If it does, we'll close the older one.

DatabasKarlsson:
Well, there were so many things to correct in the old one, so please use 
this one instead. I'm not really going to touch this again, it's up to 
you if you want to include it.

/Karlsson

markus456:
If you want us to address the review comments, please enable commits from maintainers: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #218, opened 2020-12-16
MXS-3342: Persistent connections cause signal 11 when no matching con…

dicode-nl:
…nection is found

There is no guarantee that the persistent pool contains a connection which matches our host
when proxy protocol is un use. We should handle this.

license.

license.


markus456 (approved):
Yup, this will fix it.

jhnwkmn (approved):

PR #227, opened 2021-05-19
Fix memory leak for explain

Daniel-Xu:
For SQL: describle table_name

The reduced srclist is not freed which will cause memory leak.

license.

### DIFF
--- a/query_classifier/qc_sqlite/qc_sqlite.cc
+++ b/query_classifier/qc_sqlite/qc_sqlite.cc
@@ -2400,6 +2400,8 @@ public:
                 update_names(pList->a[i].zDatabase, pList->a[i].zName, pList->a[i].zAlias, nullptr);
             }
         }
+
+        exposed_sqlite3SrcListDelete(pParse->db, pList);
     }
 
     void maxscaleExplain(Parse* pParse)


query_classifier/qc_sqlite/sqlite-src-3110100/src/parse.y
```diff
@@ -307,7 +307,7 @@ ecmd ::= oracle_assignment SEMI.
 explain_kw ::= EXPLAIN.  // Also covers DESCRIBE
 explain_kw ::= DESC.
 
-explain ::= explain_kw tbl_name(A). { pParse->explain = 1; maxscaleExplainTable(pParse, A); }
+explain ::= explain_kw tbl_name(A). { pParse->explain = 1; maxscaleExplainTable(pParse, A); sqlite3SrcListDelete(pParse->db, A); }
```
jhnwkmn: But for some exceptions, the data structures that are passed to the qc_sqlite side are also freed on that side. So please add instead at the end of the maxscaleExplainTable()

    exposed_sqlite3SrcListDelete(pParse->db, pList);

jhnwkmn (changes requested):

query_classifier/qc_sqlite/sqlite-src-3110100/src/parse.y
Daniel-Xu: Sure. I've tested it and it worked

jhnwkmn (approved):

PR #230, opened 2021-07-11
support mysql8.0.x

happy-v587:
support 8.0.x

### DIFF
--- a/server/core/service.cc
+++ b/server/core/service.cc
@@ -144,7 +144,7 @@ static std::string get_version_string(mxs::ConfigParameters* params)
 {
     std::string version_string = params->get_string(CN_VERSION_STRING);
 
-    if (!version_string.empty() && version_string[0] != '5')
+    if (!version_string.empty() && version_string[0] != '5' && version_string[0] != '8')
     {
         /**
          * Add the 5.5.5- string to the start of the version string if the version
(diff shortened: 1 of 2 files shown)


markus456:
This should be done in a more controllable manner so that you can use MySQL connectors with MariaDB. Adding it as a configuration parameter would allow it to be enabled and disabled. Another option might be to see if there's a way to detect that the client doesn't expect the `5.5.5-` prefix and drop it only for those clients.

happy-v587:
Thank you for your reply. Later I will change it as configuration parameter, so that can control it.

markus456:
I think I have to apologize: I misunderstood which version_string this change affected. I thought it was for the one returned by the backend servers themselves that MaxScale uses by default but instead it is for the `version_string` parameter defined in the configuration file.

Now that I've taken a second look at this, I think just the original commit would be adequate as the new parameter doesn't seem to do that much.

markus456:
Additionally, we require that any code contributions to MaxScale are submitted under the BSD-new license. Please indicate this by leaving a comment with the following text:

> I am contributing the new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license.

happy-v587:
I am contributing the new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license.

markus456 (commented):
I mean that the original commit that only checks whether the `version_string` starts with an 8 would have been enough. Can you use that commit?

happy-v587:
Ok, I've revised it.

markus456 (approved):

mbnantti (approved):

PR #254, opened 2022-04-27
add Gui demo instruction

timofeyturenko:
license.


system-test/gui_demo.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+using namespace std;
+
+int main(int argc, char* argv[])
+{
+    TestConnections* Test = new TestConnections(argc, argv);
+    Test->start_maxscale(0);
+    int rval = Test->global_result;
+    delete Test;
+    return rval;
```
markus456: This could just be:
```c++
TestConnections test(argc, argv);
return test.global_result;
```

system-test/gui_demo.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <iostream>
+#include <maxtest/testconnections.hh>
+
+using namespace std;
+
+int main(int argc, char* argv[])
+{
+    TestConnections test(argc, argv);
+    test.start_maxscale(0);
```
markus456: This shouldn't be needed, the test always starts MaxScale.

system-test/gui_demo.cpp
```diff
@@ -0,0 +1,16 @@
+/**
+ * @file gui_demo.cpp Dummy test to start Maxscale GUI demo
+ */
+
+
+#include <iostream>
```
markus456: This isn't needed as it's not used.

system-test/gui_demo.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <maxtest/testconnections.hh>
+
+using namespace std;
```
markus456: This is also not needed as it's not used.

markus456 (approved):
Looks good, minor changes could be made but they aren't that important.

PR #262, opened 2022-08-11
fix complie error for static_assert

yezhiweicmss:
license.


jhnwkmn:
MaxScale 6 is build using C++17 that supports single argument static_assert.

https://en.cppreference.com/w/cpp/language/static_assert

Are you using C++17?

yezhiweicmss:
Got it.

PR #354, opened 2024-02-06
Fix Galera cluster creation for system-test

timofeyturenko:
now galera_server*.cnf files are configured by `galera_config` MDBCI product, but these files do not contain `wsrep_cluster_address`. This parameter is configured by system-test code in the separate cluster_address.cnf file. It means after MDBCI provisioning but before creation of cluster_address.cnf by system-test the server configuration is broken. The Community Server ignores it and the server can be started, but the Enterprise server fails to start.
The solution is to remove `galera_config` product and move galera_server*.cnf files coping and configuration to system-test code.


markus456:
Please rebase this on top of 6.4.

PR #355, opened 2024-02-07
Fix Galera cluster creation for system-test

timofeyturenko:
now galera_server*.cnf files are configured by galera_config MDBCI product, but these files do not contain wsrep_cluster_address. This parameter is configured by system-test code in the separate cluster_address.cnf file. It means after MDBCI provisioning but before creation of cluster_address.cnf by system-test the server configuration is broken. The Community Server ignores it and the server can be started, but the Enterprise server fails to start.
The solution is to remove galera_config product and move galera_server*.cnf files coping and configuration to system-test code.

### DIFF
--- a/system-test/maxtest/src/galera_cluster.cc
+++ b/system-test/maxtest/src/galera_cluster.cc
@@ -48,12 +48,6 @@ bool GaleraCluster::start_replication()
     auto gcomm = ss.str();
 
     run_on_every_backend([&](int i){
-        copy_to_node(i,
-                     (mxt::SOURCE_DIR + std::string("/mdbci/cnf/") + get_srv_cnf_filename(i)).c_str(),
-                     access_homedir(i));
-        ssh_node(i,
-                 (std::string("cp ") + get_srv_cnf_filename(i) + std::string(" /etc/my.cnf.d/")).c_str(),
-                  true);
         ssh_node(i, "echo [mysqld] > cluster_address.cnf", true);
         ssh_node_f(i, true, "echo wsrep_cluster_address=gcomm://%s >>  cluster_address.cnf", gcomm.c_str());
         ssh_node(i, "cp cluster_address.cnf /etc/my.cnf.d/", true);
--- a/system-test/maxtest/src/galera_cluster.cc
+++ b/system-test/maxtest/src/galera_cluster.cc
@@ -49,7 +49,7 @@ bool GaleraCluster::start_replication()
 
     run_on_every_backend([&](int i){
         copy_to_node(i,
-                     (m_shared.vm_path + std::string("/") + get_srv_cnf_filename(i)).c_str(),
+                     (mxt::SOURCE_DIR + std::string("/mdbci/cnf/") + get_srv_cnf_filename(i)).c_str(),
                      access_homedir(i));
         ssh_node(i,
                  (std::string("cp ") + get_srv_cnf_filename(i) + std::string(" /etc/my.cnf.d/")).c_str(),
@@ -60,9 +60,9 @@ bool GaleraCluster::start_replication()
         ssh_node(i, "cp cluster_address.cnf /etc/mysql/my.cnf.d/", true);
 
         ssh_node(i, "rm -rf /var/lib/mysql/*", true);
-        unblock_node(i, 4567);
-        unblock_node(i, 4568);
-        unblock_node(i, 4444);
+        unblock_node_port(i, 4567);
+        unblock_node_port(i, 4568);
+        unblock_node_port(i, 4444);
         ssh_node(i, "mariadb-install-db --user=mysql", true);
 
         ssh_node_f(i,
--- a/system-test/maxtest/src/galera_cluster.cc
+++ b/system-test/maxtest/src/galera_cluster.cc
@@ -48,12 +48,21 @@ bool GaleraCluster::start_replication()
     auto gcomm = ss.str();
 
     run_on_every_backend([&](int i){
+        copy_to_node(i,
+                     (m_shared.vm_path + std::string("/") + get_srv_cnf_filename(i)).c_str(),
+                     access_homedir(i));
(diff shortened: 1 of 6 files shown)


system-test/maxtest/src/mariadb_nodes.cc
```diff
@@ -414,6 +414,21 @@ std::string MariaDBCluster::unblock_command(int node) const
     return command;
 }
 
+std::string MariaDBCluster::unblock_port_command(int port) const
```
markus456, line 407: The `unblock_command` should be changed to:
```c++
std::string MariaDBCluster::unblock_command(int node) const
{
    return unblock_port_command(port[node]);
}
```

This way you don't have nearly identical code in two functions.

system-test/maxtest/src/mariadb_nodes.cc
```diff
@@ -431,6 +446,13 @@ bool MariaDBCluster::unblock_node(int node)
     return res == 0;
 }
 
+bool MariaDBCluster::unblock_node(int node, int port)
```
markus456: This should not be an overload of `unblock_node()` since it has different semantics than the original `unblock_node()`. The new one doesn't call `clean_iptables()` and if that's not required, it should be named `unblock_node_port()` instead.

For `std::string` commands, there's also no need to use `ssh_node_f` as there's no formatting that's done. You can simply implement it as:
```c++
bool MariaDBCluster::unblock_node(int node, int port)
{
    return ssh_node(unblock_port_command(port), true);
}
```

system-test/maxtest/src/galera_cluster.cc
```diff
@@ -48,12 +48,21 @@ bool GaleraCluster::start_replication()
     auto gcomm = ss.str();
 
     run_on_every_backend([&](int i){
+        copy_to_node(i,
+                     (m_shared.vm_path + std::string("/") + get_srv_cnf_filename(i)).c_str(),
```
markus456: This seems wrong: `vm_path` refers to the location where the virtual machines are stored, not the location where the configuration files are. For me, that would be `~/vms/testing-vms/galera_server1.cnf` which does not exist.

And why does this need to be copied again? It's already copied to the node in `MariaDBCluster::reset_server_settings()` which is called by `GaleraCluster::reset_server()` which is a function that gets called very early in the node rebuild step.

system-test/maxtest/src/galera_cluster.cc
```diff
@@ -48,12 +48,21 @@ bool GaleraCluster::start_replication()
...
+        ssh_node(i,
+                 (std::string("cp ") + get_srv_cnf_filename(i) + std::string(" /etc/my.cnf.d/")).c_str(),
+                  true);
         ssh_node(i, "echo [mysqld] > cluster_address.cnf", true);
         ssh_node_f(i, true, "echo wsrep_cluster_address=gcomm://%s >>  cluster_address.cnf", gcomm.c_str());
         ssh_node(i, "cp cluster_address.cnf /etc/my.cnf.d/", true);
         ssh_node(i, "cp cluster_address.cnf /etc/mysql/my.cnf.d/", true);
 
         ssh_node(i, "rm -rf /var/lib/mysql/*", true);
+        unblock_node(i, 4567);
```
markus456: Some comments on what these are used for wouldn't hurt. The explanations are here: https://galeracluster.com/library/documentation/firewall-settings.html

system-test/mdbci/templates/default.json.template
```diff
@@ -86,13 +86,8 @@
...
         "version": "${version}",
         "force_version": ${force_backend_version}
-      },
-      {
-        "name": "galera_config",
-        "cnf_template" : "galera_server1.cnf"
       }
-    ],
-    "cnf_template_path": "${cnf_path}"
+    ]
```
markus456, line 82: Why do the normal MariaDB nodes require the `cnf_template` arguments but the galera ones do not? Shouldn't they all be removed if they're copied over by the test code?

markus456 (changes requested):

system-test/maxtest/src/galera_cluster.cc
```diff
@@ -48,12 +48,21 @@ bool GaleraCluster::start_replication()
     auto gcomm = ss.str();
 
     run_on_every_backend([&](int i){
+        copy_to_node(i,
+                     (mxt::SOURCE_DIR + std::string("/mdbci/cnf/") + get_srv_cnf_filename(i)).c_str(),
```
markus456: We have this:
```c++
void MariaDBCluster::reset_server_settings(int node)
{
    string cnf_dir = m_test_dir + "/mdbci/cnf/";
    string cnf_file = get_srv_cnf_filename(node);
    string cnf_path = cnf_dir + cnf_file;
    string home = access_homedir(node);

    // Note: This is a CentOS specific path
    ssh_node(node, "rm -rf /etc/my.cnf.d/*", true);

    copy_to_node(node, cnf_path.c_str(), home.c_str());
    ssh_node_f(node, false, "sudo install -o root -g root -m 0644 ~/%s /etc/my.cnf.d/", cnf_file.c_str());
```
Ths is called by `reset_servers()` which is called right before `start_replication()`;
```c+
            if (reset_servers())
            {
                log.log_msgf("%s reset. Starting replication.", namec);
                start_replication();
```
Why does the file have to be copied again in `start_replication()`?

system-test/maxtest/src/mariadb_nodes.cc
```diff
@@ -400,6 +400,11 @@ std::string MariaDBCluster::block_command(int node) const
 }
 
 std::string MariaDBCluster::unblock_command(int node) const
+{
+    return  unblock_port_command(port[node]);;
```
markus456: There's an extra trailing semicolon here.

markus456 (changes requested):

PR #395, opened 2024-10-10
add ppc64 support to build scripts

timofeyturenko:
license.

### DIFF
--- a/BUILD/build_package.sh
+++ b/BUILD/build_package.sh
@@ -4,21 +4,27 @@
 
 set -x
 
+function is_ppc64() {
+    [ "$(arch)" == "ppc64le" ]
+}
+
+
 cd ./MaxScale || exit 1
 git submodule update --init
 cd ..
 
 NCPU=$(grep -c processor /proc/cpuinfo)
+NCPUSTR="-j${NCPU}"
 
-if [ "$PARALLEL_BUILD" == "no" ]
+if [ "$PARALLEL_BUILD" == "no" ] || [ is_ppc64 ]
 then
-    NCPU=1
+    NCPUSTR=""
 fi
 
 mkdir _build
 cd _build || exit 1
 cmake ../MaxScale -DCMAKE_COLOR_MAKEFILE=N $cmake_flags
-make "-j${NCPU}" || exit 1
+make ${NCPUSTR} || exit 1
 
 if [[ "$cmake_flags" =~ "BUILD_TESTS=Y" ]]
 then
@@ -29,7 +35,7 @@ then
     ctest --timeout 120 --output-on-failure "-j${NCPU}" || exit 1
 fi
 
-sudo make "-j${NCPU}" package
+sudo make ${NCPUSTR} package
 res=$?
 if [ $res != 0 ] ; then
 	echo "Make package failed"
@@ -40,4 +46,4 @@ sudo rm CMakeCache.txt
 
 echo "Building tarball..."
 cmake ../MaxScale $cmake_flags -DTARBALL=Y
-sudo make "-j${NCPU}" package
+sudo make ${NCPUSTR} package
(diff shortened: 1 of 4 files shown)


BUILD/build_package.sh
```diff
@@ -4,21 +4,27 @@
...
+
 cd ./MaxScale || exit 1
 git submodule update --init
 cd ..
 
 NCPU=$(grep -c processor /proc/cpuinfo)
+NCPUSTR="-j${NCPU}"
 
-if [ "$PARALLEL_BUILD" == "no" ]
+if [ "$PARALLEL_BUILD" == "no" ] || [ is_ppc64 ]
```
markus456, line 19: These changes should not be needed anymore now that commit 47150638f5f0ed4b1495ae47b0a481178cf36f05 has been merged into 21.06.

BUILD/install_test_build_deps.sh
```diff
@@ -24,15 +30,13 @@ then
   ${apt_cmd} update
 
   ${apt_cmd} install curl
-  curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | sudo bash
```
markus456, line 27: This has obviously worked for Ubuntu for a long time so I'd be hesitant to change it. Having to install one part of the dependencies with MDBCI and another with this script is going to make it more complex to debug.

BUILD/install_test_build_deps.sh
```diff
@@ -24,15 +30,13 @@ then
...
 
   ${apt_cmd} install \
        git wget build-essential libssl-dev \
-       mariadb-client mariadb-plugin-gssapi-client \
```
markus456, line 31: These packages were coming from the Ubuntu repositories which is why the names do not work. You need to use `MariaDB-client` to install from the MariaDB repositories.

BUILD/install_test_build_deps.sh
```diff
@@ -90,29 +86,40 @@ EOL
...
-    sudo yum install -y --nogpgcheck java-1.8.0-openjdk
-    sudo yum install -y --nogpgcheck centos-release-scl
-    sudo yum install -y --nogpgcheck devtoolset-7-gcc*
-    sudo yum install -y --nogpgcheck php-mysql
+    sudo yum install -y --nogpgcheck ${enable_power_tools} ${enable_crb} java-1.8.0-openjdk
+    sudo yum install -y --nogpgcheck ${enable_power_tools} ${enable_crb} centos-release-scl
+    sudo yum install -y --nogpgcheck ${enable_power_tools} ${enable_crb} devtoolset-7-gcc*
+    sudo yum install -y --nogpgcheck ${enable_power_tools} ${enable_crb} php-mysql
+    sudo yum install -y --nogpgcheck ${enable_power_tools} ${enable_crb} php-mysqlnd
     echo "please run 'scl enable devtoolset-7 bash' to enable new gcc!!"
```
markus456, line 123: I'm fairly certain that not all of these are needed. For example, the `devtoolset-7-gcc` was only needed for CentOS 7 which is EOL. Please figure out which ones are needed on RHEL 8 and RHEL 9 and add the ones that exist on both into the same `yum` command. No point in copy-pasting the same command if it can be added to the existing one.

If some of them only work on RHEL 9 and not on RHEL 8, then put them into their own `if` scope similar to how they're handled in `install_build_deps.sh`.

system-test/mdbci/run_test.sh
```diff
@@ -86,7 +91,14 @@ mkdir build && cd build
...
 NCPU=$(grep -c processor /proc/cpuinfo)
-make -j $NCPU
+
+if [ is_ppc64 ]
+then
+    make
+else
+    make -j $NCPU
+fi
+
```
markus456, line 101: These are also no longer needed.

markus456 (changes requested):

markus456:
I'll take over this task and continue working on this.

PR #400, opened 2024-11-10
Fix building deb package

kazan417:
Fixes building of debian package on distribution where codename have "_" symbols which is not allowed in package version.

license.


kazan417:
This pull request fixes such building error while running make package:
Run CPack packaging tool...
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: MaxScale
CPack: - Install project: MaxScale []
CPack: -   Install component: core
CPack: Create package
CPackDeb: - Generating dependency list
CMake Error at /usr/share/cmake-3.25/Modules/Internal/CPack/CPackDeb.cmake:407 (message):
  CPackDeb: Debian package version must confirm to "^[0-9][A-Za-z0-9.+~-]*$"
  regex!
Call Stack (most recent call first):
  /usr/share/cmake-3.25/Modules/Internal/CPack/CPackDeb.cmake:801 (cpack_deb_prepare_package_vars)


CPack Error: Error while execution CPackDeb.cmake
CPack Error: Problem compressing the directory
CPack Error: Error when generating package: maxscale
make: *** [Makefile:71: package] Error 1

markus456:
What version of Debian does this fail with?

kazan417:
> What version of Debian does this fail with?

This error appears  not in Debian distro, but on other Debian-based distributive. In Debian there is no distro containing "_" symbol in the codename now, but it is possible in the future. But Debian-based package manager used not only on Debian.

This fix adds general compatibility and conformance to debian policy version numbering - 5.6.12

The version number of a package. The format is: [epoch:]upstream_version[-debian_revision]
The upstream_version must contain only alphanumerics  and the characters . + - ~ (full stop, plus, hyphen, tilde) and should start with a digit. If there is no debian_revision then hyphens are not allowed.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#special-version-conventions

kazan417:
Reopened this by oter pull request https://github.com/mariadb-corporation/MaxScale/pull/401

PR #401, opened 2024-11-13
Fix building deb package

kazan417:
license.


kazan417:
Thi is continuation of https://github.com/mariadb-corporation/MaxScale/pull/400 because of deleting branch.

kazan417:
This fix adds general compatibility and conformance to debian policy version numbering - 5.6.12

The version number of a package. The format is: [epoch:]upstream_version[-debian_revision]
The upstream_version must contain only alphanumerics and the characters . + - ~ (full stop, plus, hyphen, tilde) and should start with a digit. If there is no debian_revision then hyphens are not allowed.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#special-version-conventions

